Issue #386, opened 2026-04-03
Vision: ScoringPlayerAgent hardcodes engine bonus constants — will silently diverge if rules change


decay256:
## Concern
ScoringPlayerAgent (#347) spec hardcodes game-engine constants in its scoring formula: tell bonus = 2, callback bonus = 1/2/3 (via gap distance), momentum bonus = 2/3 (via streak thresholds). These same values are defined in `Pinder.Core` as:
- `GameSession` line 434: `tellBonus = 2`
- `CallbackBonus.Compute()`: returns 0/1/2/3
- `GameSession.GetMomentumBonus()`: returns 0/2/3

The scoring agent duplicates these constants rather than calling the engine's own methods. If any bonus value changes in a future sprint, the scoring agent will silently produce incorrect probability estimates.

## Why it matters
The scoring agent is the deterministic foundation — `LlmPlayerAgent` (#348) falls back to it on failure. If the EV formula drifts from reality, every sim run produces misleading balance data. At prototype this is tolerable, but it creates hidden tech debt.

## What needs to happen
The scoring agent should call `CallbackBonus.Compute()` directly (it's a public static method in `Pinder.Core.Conversation`). For momentum, either expose `GetMomentumBonus` as public static, or duplicate with a TODO. The tell bonus (literal `2`) has no public constant to reference — acceptable to hardcode with a comment.

## Acceptance Criteria
- [ ] `ScoringPlayerAgent` calls `CallbackBonus.Compute(currentTurn, callbackTurnNumber)` instead of reimplementing the gap→bonus logic
- [ ] Momentum bonus logic either calls a shared method or has a `// SYNC: GameSession.GetMomentumBonus()` comment documenting the duplication
- [ ] Tell bonus hardcoded `2` has a comment referencing `GameSession ResolveTurnAsync tellBonus`

**Role**: backend-engineer
**Maturity**: prototype

Issue #347, opened 2026-04-03
Player agent: ScoringPlayerAgent — mechanical expected-value scoring


decay256:
## Context
A deterministic player agent that scores options using the game's mechanical rules. No LLM — pure math. Produces consistent, explainable decisions useful for regression testing.

## Scoring formula

For each option, compute:
```
expectedInterestGain = successChance × (baseInterestGain + riskTierBonus + hiddenBonuses)
                     - failChance × failCost

where:
  successChance = max(0, min(1, (21 - need) / 20))
  need = DC - (statMod + momentumBonus + callbackBonus + tellBonus)
  baseInterestGain = 1 if margin 1-4, 2 if 5-9, 3 if 10+
  riskTierBonus = 0 (Safe/Medium), +1 (Hard), +2 (Bold)
  hiddenBonuses = callbackBonus (applied to roll, affects tier) + comboBonus (interest)
  failCost = failTierInterestDelta + trapPenaltyEstimate
  trapPenaltyEstimate = if trap would fire (miss 6-9): ~-2 interest per turn for trap duration
```

## Strategic adjustments
- If momentum streak = 2: bias toward success to reach Momentum active (+2 next roll)
- If interest 19-20: pick lowest-variance option (Safe/Medium wins the match)
- If interest 5-8 (Bored, disadvantage state): favour bold (nothing to lose, trap risk is low priority)
- If trap active on this stat: heavy penalty to options using that stat

## Outputs the reasoning
Each decision should produce a `PlayerDecision` with:
- Chosen option index
- Score breakdown for all 4 options
- The strategic reason: e.g. 'Charm at 45% beats Honesty at 30% — 15pp advantage outweighs DC risk. Momentum at 2 — prioritizing success to reach +2 bonus.'

## Acceptance Criteria
- [ ] `ScoringPlayerAgent` implements `IPlayerAgent`
- [ ] Scores all 4 options per the formula above
- [ ] Applies momentum, interest state, and trap adjustments
- [ ] `PlayerDecision.Reasoning` string explains the pick
- [ ] Deterministic — same input always produces same output
- [ ] Unit tests: high-momentum state prefers safe option; Bored state prefers bold; active trap penalizes that stat
- [ ] Build clean

Depends on: #346

Depends on: #346

decay256:
## Work Summary — Code Review

**Agent**: code-reviewer
**Correlation ID**: doc-review-sprint-decay256-pinder-core-1-1775257091167-issue-347-r1
[...]

PR #394, opened 2026-04-03
[Issue #347] Write spec document: Player agent: ScoringPlayerAgent — mechanical expected-value scoring

decay256:
Refs #347

## Summary

Updated specification document for ScoringPlayerAgent — the deterministic expected-value scoring player agent.

### Key updates:
- **#386 ADR compliance**: Mandates calling `CallbackBonus.Compute()` directly (not reimplemented), momentum bonus duplicated with `// SYNC:` comment, tell bonus hardcoded `2` with `// SYNC:` comment
- Clarified stat-to-trap-name mapping table for active trap detection
- Added both simple and exact EV computation options (either acceptable at prototype maturity)
- Expanded edge cases and boundary value documentation
- Full input/output examples with worked computations

## Spec location
`docs/specs/issue-347-spec.md`

## DoD Evidence
**Branch:** issue-347-write-spec-document-player-agent-scoring
**Commit:** 448a149


decay256 (commented):
## Code Review

**Maturity level**: prototype (spec document)
**Errors (must fix)**: 0
**Warnings (must fix)**: 0

| # | Location | Note |
|---|----------|------|
| 1 | `docs/specs/issue-347-spec.md` Step 4 | Option A vs Option B for EV computation — consider noting which approach the tests will validate against, since exact weighted EV (Option B) produces different `baseInterestGain` values than the midpoint approximation (Option A). The spec says tests shouldn't validate exact EV precision, which handles this, but implementers may benefit from knowing the expected default. |

### Positive notes
- Excellent improvement: mandating `CallbackBonus.Compute()` call instead of reimplementing the logic eliminates a class of drift bugs
- The explicit `StatType → Shadow Stat` mapping table removes all ambiguity for the implementer
- SYNC comment pattern is a pragmatic approach to flagging cross-boundary duplication at prototype maturity
- New Example 4 (Bored state) and expanded edge cases (boundary values, clamping) significantly improve spec completeness
- Removing `RiskTier` from dependencies while adding `CallbackBonus` correctly reflects the actual dependency graph

**VERDICT: APPROVE** (posted as comment due to own-PR restriction — 0 errors, 0 warnings)

PR #395, opened 2026-04-03
[Issue #354] Bug: Session runner file counter writes session-001 when session-005 already exists

### LINKED ISSUE #354: Bug: Session runner file counter writes session-001 when session-005 already exists
## Context
The session runner counter logic should find the highest existing session number and increment it. Session 006 was run but wrote `session-001-sable-vs-brick.md` even though `session-005-sable-vs-brick.md` already existed.

## Root cause
The counter glob is `session-???.md` but existing files are named `session-005-sable-vs-brick.md`. The glob matches the filename, but `Path.GetFileNameWithoutExtension` gives `session-005-sable-vs-brick`, and `n.Substring(8, 3)` extracts `005-` — the 4th char is a hyphen, so `int.TryParse("005-", ...)` fails, returning 0. Counter stays at 1.

## Fix
Change the filename extraction to handle the full name format:
```csharp
// Instead of Substring(8, 3), parse up to the next hyphen after 'session-'
var name = Path.GetFileNameWithoutExtension(f);
// name = "session-005-sable-vs-brick"
// Extract the number after "session-"
var parts = name.Split('-');
```

## Acceptance Criteria
- [ ] Counter correctly reads existing session file numbers
- [ ] Running after session-005 produces session-006, not session-001
- [ ] Works with any character name in the filename (hyphens in names are fine)
- [ ] Build clean

decay256:
Fixes #354

## Changes

- Extracted file counter logic from `Program.WritePlaytestLog` into `SessionFileCounter.GetNextSessionNumber()` (internal static method in session-runner)
- `Program.cs` now calls the extracted method instead of inlining the logic
- Tests call the **real production method** instead of reimplementing logic locally (addresses code review feedback from PR #371)
- Added `InternalsVisibleTo` on session-runner.csproj so test project can access internal types
- Glob pattern `session-*.md` and `Split('-')[1]` parsing correctly handles filenames with character name suffixes

## How to test

```bash
dotnet test tests/Pinder.Core.Tests --filter SessionFileCounter
```

## DoD Evidence
**Branch:** issue-354-bug-session-runner-file-counter-writes-s
**Commit:** 8a8e5ac
**Tests:** 5 passed (SessionFileCounter), 1505 total passed, 0 failed
**Deviations from contract:** none

### DIFF
--- a/session-runner/Program.cs
+++ b/session-runner/Program.cs
@@ -385,13 +385,7 @@ class Program
     {
         string dir = "/root/.openclaw/agents-extra/pinder/design/playtests";
         if (!Directory.Exists(dir)) { Console.Error.WriteLine("Playtest dir not found"); return; }
-        int nextNum = 1;
-        foreach (var f in Directory.GetFiles(dir, "session-*.md")) {
-            var n = Path.GetFileNameWithoutExtension(f);
-            // name = "session-005-sable-vs-brick" → Split('-') = ["session","005","sable","vs","brick"]
-            var parts = n.Split('-');
-            if (parts.Length >= 2 && int.TryParse(parts[1], out int num)) nextNum = Math.Max(nextNum, num + 1);
-        }
+        int nextNum = SessionFileCounter.GetNextSessionNumber(dir);
         string slug = $"session-{nextNum:D3}-{p1.ToLower()}-vs-{p2.ToLower()}.md";
         File.WriteAllText(Path.Combine(dir, slug), content);
         Console.WriteLine($"\n📝 Written → design/playtests/{slug}");
--- a/session-runner/SessionFileCounter.cs
+++ b/session-runner/SessionFileCounter.cs
@@ -1,0 +1,29 @@
+using System;
+using System.IO;
+
+/// <summary>
+/// Extracts the next session number from a directory of session markdown files.
+/// Files are expected to follow the naming convention: session-NNN-name-vs-name.md
+/// </summary>
+internal static class SessionFileCounter
+{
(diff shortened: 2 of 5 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 0
**Warnings (must fix)**: 0

| # | Location | Note |
|---|----------|------|
| 1 | `tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj` | Test project `Pinder.Core.Tests` now references `session-runner` (an Exe project). Consider creating a dedicated `SessionRunner.Tests` project in a future sprint if session-runner grows more testable surface area. Not blocking — works fine for now. |

### Positive notes
- **Key fix**: Tests previously contained a local copy of the production logic (the `GetNextSessionNumber` static method was duplicated in the test file). This is the classic "test reimplements code it's testing" anti-pattern — the test would never catch a regression because it wasn't testing the real code. Now the tests properly call `SessionFileCounter.GetNextSessionNumber` from the source assembly via `InternalsVisibleTo`.
- Clean single-responsibility extraction — `SessionFileCounter` does one thing with clear XML doc comments.
- Logic is identical pre/post refactor (verified by diffing the extracted method body against the removed inline code).
- All 5 existing tests updated to call the real method. Good coverage: empty dir, single file, multiple files, hyphenated names, non-session files ignored.
- Refactoring is properly scoped to the bug fix (no unrelated changes).

**VERDICT: APPROVE** (posted as comment due to GitHub self-review restriction — no errors or warnings found, PR is approved)

PR #411, opened 2026-04-04
[Issue #386] Write tests: Vision concern #386

### LINKED ISSUE #386: Vision: ScoringPlayerAgent hardcodes engine bonus constants — will silently diverge if rules change

decay256:
Fixes #386

## What was added

19 sync-verification tests in `BonusConstantSyncTests.cs` that guard against silent drift between `ScoringPlayerAgent` bonus calculations and the engine's canonical values.

### Test coverage:
- **CallbackBonus.Compute() all tiers**: gap<2 (→0), gap 2-3 non-opener (→1), gap≥4 non-opener (→2), opener (→3)
- **Agent applies correct tier per gap**: verifies success chance delta between opener (+3) and mid-distance (+1)
- **Momentum boundary values**: streak 2 (no bonus), streak 3 (+2), streak 4 (still +2), streak 5 (+3)
- **Tell bonus exact +2**: verified via d20 success chance delta of 0.10
- **Additive stacking**: all bonuses combine correctly (momentum+tell+callback = +7)
- **BonusesApplied completeness**: all active bonuses appear in the list
- **No callback → no bonus**: null CallbackTurnNumber yields no callback entry

Each test includes a mutation comment explaining what specific drift it would catch.

## DoD Evidence
**Branch:** issue-386-write-tests-vision-concern-386
**Commit:** 3060688
**Tests:** All 19 pass (`dotnet test --filter BonusConstantSync`)

### DIFF
--- a/tests/Pinder.Core.Tests/BonusConstantSyncTests.cs
+++ b/tests/Pinder.Core.Tests/BonusConstantSyncTests.cs
@@ -1,0 +1,352 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Pinder.Core.Conversation;
+using Pinder.Core.Stats;
+using Pinder.SessionRunner;
+using Xunit;
+
+namespace Pinder.Core.Tests
+{
+    /// <summary>
+    /// Issue #386: Sync-verification tests ensuring ScoringPlayerAgent bonus constants
+    /// match the engine's canonical values. These tests guard against silent drift
+    /// between the session-runner scoring agent and Pinder.Core game rules.
+    /// </summary>
+    public class BonusConstantSyncTests
+    {
+        private readonly ScoringPlayerAgent _agent = new ScoringPlayerAgent();
+
+        #region Helpers
+
+        private static StatBlock MakeStats(
+            int charm = 0, int rizz = 0, int honesty = 0,
+            int chaos = 0, int wit = 0, int sa = 0,
+            int madness = 0, int horniness = 0, int denial = 0,
+            int fixation = 0, int dread = 0, int overthinking = 0)
+        {
+            return new StatBlock(
+                new Dictionary<StatType, int>
+                {
+                    { StatType.Charm, charm },
+                    { StatType.Rizz, rizz },
+                    { StatType.Honesty, honesty },
+                    { StatType.Chaos, chaos },
+                    { StatType.Wit, wit },
+                    { StatType.SelfAwareness, sa }
+                },
+                new Dictionary<ShadowStatType, int>
+                {
+                    { ShadowStatType.Madness, madness },
+                    { ShadowStatType.Horniness, horniness },
+                    { ShadowStatType.Denial, denial },
+                    { ShadowStatType.Fixation, fixation },
+                    { ShadowStatType.Dread, dread },
+                    { ShadowStatType.Overthinking, overthinking }
+                });
+        }
+
+        private static DialogueOption MakeOption(
+            StatType stat,
+            int? callbackTurn = null,
+            string? comboName = null,
+            bool hasTellBonus = false)
+        {
+            return new DialogueOption(
+                stat, $"{stat} option",
+                callbackTurnNumber: callbackTurn,
+                comboName: comboName,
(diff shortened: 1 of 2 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 1
**Warnings (must fix)**: 0

### Errors

| 1 | `ScoringPlayerAgentTests.cs:500-506` (`MomentumBonus_MatchesGameSessionThresholds`) | **Import vs. Rewrite violation.** The test reimplements the momentum bonus logic locally (`if (streak >= 5) agentBonus = 3; else if (streak >= 3) agentBonus = 2; else agentBonus = 0;`) and asserts this local copy against hardcoded `[InlineData]` expected values. No production code is called — this test would pass even if the engine and agent used completely different thresholds. A test that reimplements the code it is testing catches nothing. | Either (a) call the real engine method (e.g., `GameSession.GetMomentumBonus(streak)` or equivalent) and assert its return value matches `expectedBonus`, or (b) call `_agent.DecideAsync()` with different momentum values and verify the success chance deltas match expected bonus tiers (as already done correctly in `BonusConstantSyncTests.MomentumBonus_Streak3_AppliesBonusToSuccessChance` etc.). Delete the local `if/else` rewrite. | Confirm the test method no longer contains any local momentum logic — `grep -n "agentBonus" tests/Pinder.Core.Tests/ScoringPlayerAgentTests.cs` should return nothing. The test should import/call a production method. |

| # | Location | Note |
|---|----------|------|
| 1 | `BonusConstantSyncTests.cs` + `ScoringPlayerAgentTests.cs` | There is significant overlap between the new `BonusConstantSyncTests` file and the tests added to `ScoringPlayerAgentTests`. For example, callback opener +3, tell bonus +2, and momentum tier verification are tested in both files with nearly identical setups. Consider consolidating into one location to reduce maintenance burden. |

### Positive notes
- `BonusConstantSyncTests` is well-structured — it calls `CallbackBonus.Compute()` from the real engine and verifies agent behavior via `DecideAsync()`, with precise delta assertions (e.g., `0.10f` for +2 on d20). This is the right way to do sync tests.
- Each test has a mutation comment explaining what it would catch — good practice.
- The callback tier tests (gap<2, gap 2-3, gap 4+, opener) cover all boundary conditions thoroughly.
- The stacking test (`AllBonuses_StackAdditively`) verifies a critical integration behavior.

**VERDICT: REQUEST_CHANGES**

decay256:
Closing stale PR — referenced issue is closed.

PR #431, opened 2026-04-04
[Issue #418] Write tests: Session runner: fix file counter to correctly read session numbers from existing playtest files

### LINKED ISSUE #418: Session runner: fix file counter to correctly read session numbers from existing playtest files

decay256:
Fixes #418

## DoD Evidence
**Branch:** issue-418-write-tests-session-runner-fix-file-coun
**Commit:** d1aa19a

## Test Summary
23 spec-driven tests for SessionFileCounter covering:
- **AC1**: Counter returns correct next number (sequential, gaps, 8→9)
- **AC2**: Path resolution (write-read flow, trailing slash, .. segments)
- **AC3**: Character names with digits/hyphens parse correctly
- **Edge cases**: Empty dir, non-md files, non-session files, large numbers, non-numeric parts, session-0
- **ResolvePlaytestDirectory**: env var priority, fallback on bad env var, walk-up from nested dir, null when not found, empty env var
- **Integration**: Resolved path works with GetNextSessionNumber

### DIFF
--- a/session-runner/Program.cs
+++ b/session-runner/Program.cs
@@ -406,11 +406,11 @@ class Program
 
     static void WritePlaytestLog(string content, string p1, string p2, GameOutcome? outcome, int xp, int turns)
     {
-        string dir = "/root/.openclaw/agents-extra/pinder/design/playtests";
-        if (!Directory.Exists(dir)) { Console.Error.WriteLine("Playtest dir not found"); return; }
+        string? dir = SessionFileCounter.ResolvePlaytestDirectory(AppContext.BaseDirectory);
+        if (dir == null) { Console.Error.WriteLine("Playtest dir not found — set PINDER_PLAYTESTS_PATH or ensure design/playtests/ exists"); return; }
         int nextNum = SessionFileCounter.GetNextSessionNumber(dir);
         string slug = $"session-{nextNum:D3}-{p1.ToLower()}-vs-{p2.ToLower()}.md";
         File.WriteAllText(Path.Combine(dir, slug), content);
-        Console.WriteLine($"\n📝 Written → design/playtests/{slug}");
+        Console.WriteLine($"\n📝 Written → {dir}/{slug}");
     }
 }
--- a/session-runner/SessionFileCounter.cs
+++ b/session-runner/SessionFileCounter.cs
@@ -4,15 +4,21 @@ using System.IO;
 /// <summary>
 /// Extracts the next session number from a directory of session markdown files.
 /// Files are expected to follow the naming convention: session-NNN-name-vs-name.md
+/// Also resolves the playtest directory path via environment variable or directory walking.
 /// </summary>
 internal static class SessionFileCounter
 {
     /// <summary>
+    /// Environment variable that overrides default playtest directory search paths.
+    /// </summary>
+    internal const string EnvVarName = "PINDER_PLAYTESTS_PATH";
+
+    /// <summary>
     /// Scans the given directory for session-*.md files and returns the next
     /// available session number (highest existing + 1, or 1 if none exist).
     /// </summary>
-    /// <param name="directory">Directory to scan for session files.</param>
-    /// <returns>The next session number to use.</returns>
+    /// <param name="directory">Absolute path to the directory containing session files.</param>
+    /// <returns>The next session number to use (>= 1).</returns>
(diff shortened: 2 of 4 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 1
**Warnings (must fix)**: 2

### Errors

| 1 | `SessionFileCounterSpecTests.cs` + `SessionFileCounterTests.cs` | **Shared mutable state across parallel test classes**: Both test classes set/clear `Environment.SetEnvironmentVariable(SessionFileCounter.EnvVarName, ...)` which is process-global. xUnit runs different test classes in parallel by default. If `SessionFileCounterSpecTests.Resolve_EnvVarOverride_TakesPriority` runs concurrently with `SessionFileCounterTests.ResolvePlaytestDirectory_EnvVarOverride`, they will clobber each other's env var, causing flaky failures. | Either (a) add `[Collection("EnvVarTests")]` with a shared `CollectionDefinition` to both test classes to serialize them, or (b) add an `xunit.runner.json` with `"parallelizeTestCollections": false`, or (c) consolidate all env-var-touching tests into a single class. | Run `dotnet test` 10+ times in succession — no flaky failures on the `Resolve_*` tests. |

### Warnings

| 1 | `SessionFileCounterSpecTests.cs:233` (`Resolve_NothingFound_ReturnsNullOrFallback`) | **Test asserts nothing meaningful**: `Assert.True(resolved == null \|\| Directory.Exists(resolved))` can never fail — if it returns null, passes; if it returns the hardcoded fallback (which exists on the build machine), also passes. This test cannot catch a bug where the method returns a wrong/stale path. | Assert specifically: create an isolated temp dir with no `design/playtests/` ancestor, set env var to null, and assert `resolved == null` (or if the hardcoded fallback exists on the system, assert it equals that exact path). The intent should be testable. | Review the assertion — it should be `Assert.Null(resolved)` when run in an environment without the hardcoded fallback, or the test should explicitly verify the fallback path value. |
| 2 | `SessionFileCounterTests.cs:249` (`ResolvePlaytestDirectory_ReturnsNullWhenNotFound`) | Same issue as above — `Assert.True(resolved == null \|\| Directory.Exists(resolved))` is a tautology that cannot fail. A test that cannot fail catches no mutations. | Same fix — make the assertion specific. If the hardcoded fallback is not guaranteed to exist in CI, use `Assert.Null(resolved)`. If it might exist, mock or isolate the fallback. | Grep for `resolved == null \|\| Directory.Exists` — should find zero instances after fix. |

| # | Location | Note |
|---|----------|------|
| 1 | `SessionFileCounterSpecTests.cs` + `SessionFileCounterTests.cs` | Significant test duplication between the two files — e.g., empty directory, gaps in sequence, character names with digits, large numbers, non-numeric parts, env var override, and walk-up tests are all duplicated. Consider consolidating into a single test class or clearly delineating unit vs integration scope. |
| 2 | `SessionFileCounter.cs:62` | The hardcoded fallback path (`/root/.openclaw/agents-extra/pinder/design/playtests`) couples the library to a specific deployment layout. This is acceptable at prototype maturity but should be removed or made configurable before MVP. |

### Positive notes
- Good 3-tier resolution strategy (env var → walk-up → fallback) — flexible and testable.
- Mutation comments on each test are helpful for understanding intent.
- Tests cover all spec acceptance criteria (AC1–AC4) and edge cases.
- `Dispose()` properly cleans up temp dirs and env vars.
- Implementation is clean, readable, and well-documented with XML comments.

**VERDICT: REQUEST_CHANGES**

PR #440, opened 2026-04-04
[Issue #415] Session runner: use CharacterAssembler + JSON data files instead of pre-assembled prompt files

### LINKED ISSUE #415: Session runner: use CharacterAssembler + JSON data files instead of pre-assembled prompt files
## Context
The simulation loads pre-assembled prompt files from `design/examples/` that were written by hand. This bypasses the entire character assembly pipeline (`CharacterAssembler`, item fragment loading, anatomy tier loading, stat aggregation, timing profile computation).

This means:
- `CharacterAssembler` is never tested end-to-end in simulation
- Bugs in fragment concatenation, archetype aggregation, or timing profile calculation are invisible
- The assembled prompts in `design/examples/` may drift from what the real assembler produces
- Testing a new item or anatomy combination requires manually updating example files

## What to build

The runner should accept raw character creation inputs and run them through the real pipeline:

### Option A — from item/anatomy definition file
```bash
dotnet run --project session-runner -- --player-def player-gerald.json --opponent-def opp-velvet.json
```

Where `player-gerald.json` contains:
```json
{
  "name": "Gerald_42",
  "gender_identity": "he/him",
  "bio": "Just a normal guy who loves the gym, good food, and real connections.",
  "level": 5,
  "items": ["backwards-snapback", "salmon-polo-collar-popped", "chinos-too-tight", "boat-shoes-no-socks", "apple-watch", "gym-membership-keychain"],
  "anatomy": {
    "length": "long",
    "girth": "average",
    "circumcision": "circumcised",
    "veins": "prominent",
    "texture": "natural",
    "balls": "substantial",
    "tattoos": "none",
    "eyes": "wide"
  },
  "build_points": { "charm": 6, "rizz": 5, "honesty": 2, "chaos": 4, "wit": 2, "self_awareness": 2 },
  "shadows": { "madness": 5, "denial": 2, "fixation": 4, "dread": 3, "overthinking": 2 }
}
```

The runner calls:
```csharp
var assembler = new CharacterAssembler(itemRepo, anatomyRepo);
var profile = assembler.Assemble(definition);
```

### Option B — shortcut for named starter characters
Keep `--player gerald` as shorthand that loads from a definition file in `data/characters/gerald.json`.

### CharacterAssembler integration
`CharacterAssembler.Assemble()` already exists in `src/Pinder.Core/Characters/CharacterAssembler.cs`. The runner needs to:
1. Load `JsonItemRepository` from `data/items/starter-items.json`
2. Load `JsonAnatomyRepository` from `data/anatomy/anatomy-parameters.json`
3. Call `CharacterAssembler.Assemble()` with item IDs + anatomy selections
4. Use the assembled `CharacterProfile` (including stat block and system prompt) in `GameSession`

## Acceptance Criteria
- [ ] Runner accepts `--player-def <path>` and `--opponent-def <path>` args
- [ ] `CharacterAssembler` is called with real item/anatomy data
- [ ] Stat block computed from items + anatomy + build points (not hardcoded)
- [ ] System prompt assembled from fragments (not loaded from `design/examples/`)
- [ ] Starter character definition files created at `data/characters/{name}.json` for all 5 characters
- [ ] Shorthand `--player gerald` maps to `data/characters/gerald.json`
- [ ] Build clean, all tests pass

Depends on: #414

decay256:
Fixes #415

## What was implemented
- **CharacterDefinitionLoader**: Parses character definition JSON files and runs the full CharacterAssembler + PromptBuilder pipeline to produce CharacterProfile objects for GameSession
- **DataFileLocator**: Resolves data file paths by walking up from a base directory (same pattern as TrapRegistryLoader)
- **Data files**: Copied starter-items.json and anatomy-parameters.json from external repo, created 5 starter character definitions (gerald, velvet, sable, brick, zyx)
- **Program.cs**: Added --player-def/--opponent-def CLI args; --player/--opponent shorthand now auto-resolves to data/characters/{name}.json via assembler pipeline with prompt file fallback

## How to test
```bash
dotnet test  # All 2404 tests pass (1915 + 489)
dotnet test --filter "CharacterDefinitionLoader|DataFileLocator"  # 18 new tests
```

## Acceptance Criteria Coverage
- [x] Runner accepts --player-def and --opponent-def args
- [x] CharacterAssembler called with real item/anatomy data
- [x] Stat block computed from items + anatomy + build points
- [x] System prompt assembled from fragments via PromptBuilder
- [x] 5 starter character definition files created
- [x] Shorthand --player gerald maps to data/characters/gerald.json
- [x] Build clean, all tests pass

## Deviations from contract
- Anatomy parameter keys in character JSON use actual parameter IDs from anatomy-parameters.json (e.g. `vein_definition`, `skin_texture`, `ball_size`, `eye_style`) rather than shortened names from the issue spec (e.g. `veins`, `texture`, `balls`, `eyes`). This is necessary for CharacterAssembler to resolve anatomy tiers correctly.

### DIFF
--- a/data/anatomy/anatomy-parameters.json
+++ b/data/anatomy/anatomy-parameters.json
@@ -1,0 +1,512 @@
+[
+  {
+    "id": "length",
+    "name": "Length",
+    "tiers": [
+      {
+        "id": "short",
+        "name": "Short",
+        "stat_modifiers": { "self_awareness": 1, "honesty": 1 },
+        "personality_fragment": "got to the jokes before anyone else; uses self-deprecation as both weapon and shield; has a particular wit that grew specifically in this soil",
+        "backstory_fragment": "heard every short joke by age fourteen; got very good at landing the punchline before anyone else could; discovered that being the one who tells the joke first is better than being the one the joke is about; has since found this principle applies to most things in life",
+        "texting_style_fragment": "gets to the point faster than anyone; doesn't pad; the punchline arrives early and the content justifies it",
+        "archetype_tendencies": ["The Philosopher", "The Sniper"],
+        "response_timing_modifier": {
+          "base_delay_delta_minutes": -5,
+          "delay_variance_multiplier": 0.8,
+          "dry_spell_probability_delta": 0.0,
+          "read_receipt": "neutral"
+        }
+      },
+      {
+        "id": "medium",
+        "name": "Medium",
+        "stat_modifiers": {},
+        "personality_fragment": "comfortable in their own skin in a way that reads as unusual; not performing modesty, genuinely fine; this reads as confidence to some and blankness to others",
+        "backstory_fragment": "grew up without this being a topic; no locker room trauma, no defining comparison moment that stuck, no incident that reordered things; had to build an identity around actual things rather than reactions to a physical fact; considers this lucky in retrospect",
+        "texting_style_fragment": "matches energy without effort; doesn't overcommunicate or undercommunicate; the neutral is a genuine neutral",
+        "archetype_tendencies": ["The Bio Responder"],
+        "response_timing_modifier": {
+          "base_delay_delta_minutes": 0,
+          "delay_variance_multiplier": 1.0,
+          "dry_spell_probability_delta": 0.0,
+          "read_receipt": "neutral"
+        }
+      },
+      {
+        "id": "long",
+        "name": "Long",
+        "stat_modifiers": { "rizz": 1, "charm": 1, "self_awareness": -1 },
+        "personality_fragment": "has a quiet awareness of their advantage that surfaces in unexpected ways; not a braggart — just knows; expects to be welcomed in a way that is mostly correct",
+        "backstory_fragment": "the first time someone made a specific remark about it they were sixteen and had no idea how to respond; filed it as significant; learned to respond by twenty; have been responding with decreasing effort and increasing ease ever since",
+        "texting_style_fragment": "unhurried; lets messages settle; never writes from desperation; knows the reply will come",
+        "archetype_tendencies": ["The Breadcrumber", "The Peacock"],
+        "response_timing_modifier": {
+          "base_delay_delta_minutes": 8,
+          "delay_variance_multiplier": 1.3,
+          "dry_spell_probability_delta": 0.05,
+          "read_receipt": "hides"
+        }
+      },
+      {
+        "id": "legendary",
+        "name": "Legendary",
+        "stat_modifiers": { "rizz": 2, "charm": 1, "self_awareness": -2 },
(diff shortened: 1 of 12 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 0
**Warnings (must fix)**: 2

### Warnings

| 1 | `session-runner/Program.cs:108-111` | `catch (Exception ex)` in `LoadCharacter` catches ALL exception types (including `NullReferenceException`, `OutOfMemoryException`, etc.) and silently falls back to the prompt file path. A bug in the assembler pipeline would be hidden — the user would get degraded output with no clear signal. | Catch specific exceptions only: `catch (FormatException ex)` and `catch (FileNotFoundException ex)`. Let unexpected exceptions propagate so they are visible. | Review the catch block — it should not catch `Exception`. `grep -n "catch (Exception" session-runner/Program.cs` should return zero matches. |
| 2 | `session-runner/CharacterDefinitionLoader.cs:124-126,140-141` | `ParseItemIds` silently skips non-string elements in the items array, and `ParseAnatomySelections` silently skips non-string values. If a data file has a typo (e.g. `"items": [123, "valid-item"]`), the invalid entry is dropped without warning. This hides data errors. | Either throw `FormatException` for non-string elements, or at minimum log a warning. Consistent with how `ParseBuildPoints` already throws on wrong types. | Add a test with a non-string item in the array and verify it throws `FormatException`. |

| # | Location | Note |
|---|----------|------|
| 1 | `data/anatomy/anatomy-parameters.json` (skin_tone) | `skin_tone` tiers have a different schema than all other parameters (only `visual_description`, no `stat_modifiers`/`personality_fragment`/etc.). This is likely intentional but the schema inconsistency means the consuming `IAnatomyRepository` must handle both shapes gracefully. Consider documenting which fields are optional in a schema comment or validation layer. |
| 2 | `session-runner/DataFileLocator.cs:56-65` | `FindRepoRoot` uses the presence of both `data/` and `src/` subdirectories as the heuristic for repo root. This is fragile (any directory on the system matching this pattern would be a false positive). Acceptable for a session runner tool, but worth noting if this is reused elsewhere. |
| 3 | `tests/Pinder.Core.Tests/DataFileLocatorTests.cs:49-53` | `FindRepoRoot_FromRootDir_ReturnsNull` has no assertion — the comment says "may or may not find it" and "should not throw". Add `Assert.Null(root)` or remove the test; a test without an assertion tests nothing. |

### Positive notes
- Clean separation: `CharacterDefinitionLoader` handles JSON→domain, `DataFileLocator` handles path resolution, `Program.cs` orchestrates. Each class does one thing.
- Error paths are well-tested: malformed JSON, missing fields, invalid level, unknown stat types, file not found — all covered.
- Graceful fallback from assembler pipeline to legacy prompt files is a good migration strategy.
- `Parse()` exposed as `internal` for testing without file I/O — good testability design.
- Data files are well-structured and consistent (aside from the skin_tone schema note above).

**VERDICT: REQUEST_CHANGES**

Issue #446, opened 2026-04-04
Hybrid rule engine: RuleBook + RuleEngine that evaluates YAML conditions against GameState


decay256:
## Context
70% of Pinder's rules are pure data (thresholds, ranges, deltas, mappings). These should live in YAML and be evaluated by a generic engine — not hardcoded in 15+ C# files. This enables balancing without recompilation and rule iteration with version-controlled diffs.

### RuleBook
Loads and indexes all rule entries from enriched YAML:
```csharp
public sealed class RuleBook
{
    public static RuleBook LoadFrom(string yamlPath);
    public IEnumerable<RuleEntry> GetRulesByType(string type);
    public RuleEntry? GetById(string id);
}
```

### Condition evaluator
```csharp
public static class ConditionEvaluator
{
    public static bool Evaluate(RuleCondition condition, GameState state);
    // Supports: miss_range, interest_range, shadow_threshold, stat_type, tier, etc.
}
```

### Outcome dispatcher
```csharp
public static class OutcomeDispatcher
{
    public static void Dispatch(RuleOutcome outcome, GameState state, IEffectHandler handler);
}

public interface IEffectHandler
{
    void ApplyInterestDelta(int delta);
    void ActivateTrap(string trapId);
    void ApplyShadowGrowth(ShadowStatType shadow, int delta, string reason);
    void SetRollModifier(string modifier); // advantage, disadvantage, +N
}
```

### Usage in GameSession
```csharp
// Instead of:
int delta = FailureScale.GetInterestDelta(result.Tier);

// With rule engine:
var rules = _ruleBook.GetRulesByType("interest_change");
var matchingRule = rules.FirstOrDefault(r => ConditionEvaluator.Evaluate(r.Condition, state));
if (matchingRule != null) OutcomeDispatcher.Dispatch(matchingRule.Outcome, state, this);
```

## What moves to YAML
- All failure tier → interest delta mappings (§5)
- All interest state → modifier mappings (§6)
- All shadow threshold → effect mappings (§7)
- All momentum → roll bonus mappings (§15)
- All risk tier → XP multiplier mappings (§15)
- All trap definitions (§14) — already in traps.json, extend with rule engine format

## What stays in C# code
- Combo detection logic (sequence pattern matching)
- LLM prompt construction (§3.1-3.8)
- Character assembly pipeline
- Complex conditional logic with multiple dependencies
- Any rule that needs access to the full session state in ways the DSL can't express

## Balancing workflow after this change
```bash
[...]

Issue #443, opened 2026-04-04
Rules DSL: fix remaining round-trip diffs (paragraph reordering, table formatting)


decay256:
## Context
The round-trip conversion (Markdown → YAML → Markdown) currently has 1251 remaining diff lines after the heading fix. These are not information loss but they indicate the generator doesn't perfectly reproduce the original formatting.

## Categories to fix
1. **Paragraph reordering (~25% of remaining diffs)** — multi-paragraph sections emit paragraphs in a different order relative to tables. Fix: preserve block order in extract.py by storing blocks as an ordered list, not grouping by type.
2. **Table formatting (~10%)** — column widths normalized to minimal separators. Fix: preserve original column widths in extract.py table_rows metadata.

## Goal
Round-trip diff lines < 50 per document (whitespace-only diffs are acceptable).

## Acceptance Criteria
- [ ] Paragraph reordering fixed — extracted/generated order matches original
- [ ] Table column widths preserved
- [ ] Re-run roundtrip_test.sh on all 9 docs
- [ ] All diffs < 50 lines per doc
- [ ] Report: before/after line counts

**Role**: backend-engineer
**Maturity**: prototype

PR #452, opened 2026-04-04
[Issue #446] Spec: Hybrid rule engine — RuleBook + RuleEngine evaluates YAML conditions against GameState

decay256:
Refs #446

## Summary
Specification document for the hybrid rule engine (Pinder.Rules project). Covers:
- RuleBook, ConditionEvaluator, OutcomeDispatcher, IEffectHandler, GameState, RuleEntry
- Full type signatures for all public APIs
- Input/output examples for loading, condition evaluation, and outcome dispatch
- §5 failure tier and §6 interest state equivalence criteria
- Edge cases (YAML parsing, YamlDotNet boxing, null/empty inputs)
- Error conditions (strict loading, lenient evaluation)
- Dependencies (YamlDotNet 16.3.0, enriched YAML from #443/#444)
- Architecture decision: GameSession integration deferred to preserve Pinder.Core zero-dep invariant

## Spec File
docs/specs/issue-446-spec.md

## DoD Evidence
**Branch:** issue-446-write-spec-document-hybrid-rule-engine-r
**Commit:** e321dc4

### DIFF
--- a/docs/specs/issue-446-spec.md
+++ b/docs/specs/issue-446-spec.md
@@ -1,0 +1,506 @@
+# Spec: Issue #446 — Hybrid Rule Engine: RuleBook + RuleEngine
+
+**Module**: docs/modules/rule-engine.md (create new)
+
+---
+
+## Overview
+
+This feature introduces a new `Pinder.Rules` project containing a generic, data-driven rule engine that loads enriched YAML rule definitions and evaluates their conditions against a game state snapshot. The engine replaces hardcoded numeric constants in `FailureScale` and `InterestMeter` with YAML-sourced values, enabling game-balance iteration without recompilation. At prototype maturity, the engine proves equivalence with existing hardcoded logic via tests; direct `GameSession` wiring is deferred to preserve `Pinder.Core`'s zero-dependency invariant.
+
+---
+
+## Function Signatures
+
+All types live in namespace `Pinder.Rules`. The project targets `netstandard2.0` with `LangVersion 8.0` and depends on `Pinder.Core` + `YamlDotNet 16.3.0`.
+
+### RuleEntry
+
+```csharp
+namespace Pinder.Rules
+{
+    /// <summary>
+    /// POCO representing a single rule entry deserialized from enriched YAML.
+    /// </summary>
+    public sealed class RuleEntry
+    {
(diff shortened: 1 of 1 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype (spec document)
**Errors (must fix)**: 0
**Warnings (must fix)**: 1

### Warnings

| 1 | `docs/specs/issue-446-spec.md` — Example 5, line ~283 | Outcome `starting_interest: 5` is used in the example but is **not listed in the recognized outcome keys table**. Per the spec's own rules, unknown outcome keys are silently ignored — so this example implies an effect that would never fire. An implementer reading this will either (a) think the outcome table is incomplete and try to implement `starting_interest`, or (b) be confused why the example shows a no-op outcome. Either way this will cause bugs or wasted effort. | Either add `starting_interest` to the recognized outcome keys table (with its handler call, e.g. `handler.SetStartingInterest(value)` + add the method to `IEffectHandler`), or replace the example outcome with keys that already exist in the table (e.g. `interest_delta: 5`). | Grep the outcome keys table and verify every outcome key used in examples appears in the table. |

| # | Location | Note |
|---|----------|------|
| 1 | `OutcomeDispatcher.Dispatch` signature | `GameState state` is accepted as a parameter but no recognized outcome key uses it in dispatch logic. If it's for future extensibility, consider adding a brief comment in the spec noting why it's there. Otherwise remove it to keep the API minimal. |
| 2 | Outcome keys table — `roll_bonus` | Dispatch description `handler.SetRollModifier("+{value}")` uses what looks like C# string interpolation syntax but without the `$` prefix. Clarify whether this means `$"+{value}"` (interpolated) or the literal string `+{value}`. |

### Positive notes
- Very thorough spec: clear type signatures, input/output examples, edge cases, error conditions, and dependency documentation
- The error philosophy (strict loading, lenient evaluation) is well articulated and consistent
- AC7 deferral is properly justified with rationale about the zero-dependency invariant
- Edge case coverage for YamlDotNet boxing (`int` vs `long`) is a nice catch that will save debugging time
- Clean separation: `Pinder.Rules → Pinder.Core` one-way dependency is clearly enforced

**VERDICT: REQUEST_CHANGES** (posted as comment due to GitHub own-PR limitation — this PR should NOT be merged until Warning #1 is resolved)

Issue #444, opened 2026-04-04
Rules DSL: enrich all 9 YAML files with explicit condition/outcome fields


decay256:
## Context
Currently only `rules-v3-enriched.yaml` has explicit `condition`/`outcome` fields on 63 entries. The other 8 docs (`risk-reward`, `character-construction`, `async-time`, `extensibility`, `traps`, `archetypes`, `anatomy-parameters`, `items-pool`) still have prose-only entries.

## What to do
Run the enrichment process on all 8 remaining docs, adding structured fields to all entries that contain specific numeric thresholds, ranges, or named effects.

Priority order:
1. `risk-reward-and-hidden-depth` — combos, momentum, tells, callbacks (many numeric values)
2. `async-time` — timing multipliers, energy costs, delay formulas
3. `traps` — already has JSON equivalent; enrichment should match traps.json schema
4. `archetypes` — ghost probabilities, test trigger frequencies (mostly qualitative)
5. Others

## Acceptance Criteria
- [ ] All 9 docs have enriched YAML files
- [ ] Enriched entries per doc reported
- [ ] Accuracy check run on all enriched entries
- [ ] 0 INACCURATE findings
- [ ] Total enriched entries across all docs reported

**Role**: backend-engineer
**Maturity**: prototype

Issue #445, opened 2026-04-04
Rules DSL: integrate generated test stubs into pinder-core test suite


decay256:
## Context
54 precise xUnit test stubs are in `rules/generated_tests/RulesSpecTests_Enriched.cs` but are outside the pinder-core test project. They need to be integrated so they run in CI.

## What to do
1. Copy `RulesSpecTests_Enriched.cs` into `tests/Pinder.Core.Tests/RulesSpec/`
2. Wire up the correct C# method calls — the stubs have intent (e.g. `FailureScale.GetInterestDelta(FailureTier.Misfire)`) but some method paths may need adjustment
3. Run `dotnet test` — fix any compilation errors in the stubs
4. The 17 `NotImplementedException` stubs should remain as stubs (they're for LLM-taint and qualitative effects that can't be unit-tested mechanically)
5. Add a CI note: 'these tests are auto-generated from rules/extracted/rules-v3-enriched.yaml — edit the YAML, not the .cs file'

## Auto-regeneration pipeline (bonus)
If time allows: add a `make rules` / `dotnet run tools` command that re-runs the extraction and test generation. This makes the pipeline repeatable.

## Acceptance Criteria
- [ ] 54 generated tests in `tests/Pinder.Core.Tests/RulesSpec/RulesSpecTests.cs`
- [ ] All 54 compile and pass (no failures, 17 stubs remain as skipped/NotImplemented)
- [ ] Source attribution comment at top of file
- [ ] All existing 2238 tests still pass
- [ ] Build clean

**Role**: backend-engineer
**Maturity**: prototype

PR #453, opened 2026-04-04
[Issue #444] Spec: Rules DSL — enrich all 9 YAML files with condition/outcome fields

decay256:
Refs #444

## Summary
Specification document for enriching all 9 extracted YAML rule files with explicit condition/outcome fields, enabling the Pinder.Rules engine to evaluate rules programmatically.

## Spec Location
docs/specs/issue-444-spec.md

## DoD Evidence
**Branch:** issue-444-write-spec-document-rules-dsl-enrich-all
**Commit:** 532f6e7

## Contents
- File inventory (9 YAML files, 288 total entries)
- Enrichment process and priority order
- Complete condition/outcome key vocabulary tables
- 5 concrete input/output examples (risk tiers, callback bonus, traps, timing delays, non-enrichable entry)
- Acceptance criteria mapped 1:1 from issue
- Edge cases (qualitative tables, code-only values, multi-rule entries, ID uniqueness)
- Error conditions (invalid YAML, unknown keys, type mismatches, field preservation)
- Dependencies (#443, PyYAML, traps.json cross-validation)

### DIFF
--- a/docs/specs/issue-444-spec.md
+++ b/docs/specs/issue-444-spec.md
@@ -1,0 +1,533 @@
+# Spec: Issue #444 — Rules DSL: Enrich All 9 YAML Files with Explicit Condition/Outcome Fields
+
+**Module**: docs/modules/rules-dsl.md (create new)
+
+---
+
+## Overview
+
+Currently only `rules-v3-enriched.yaml` (63 entries) has machine-readable `condition`/`outcome` fields. The other 8 extracted YAML documents (`risk-reward-and-hidden-depth`, `async-time`, `traps`, `archetypes`, `character-construction`, `items-pool`, `anatomy-parameters`, `extensibility`) contain only prose descriptions with no structured fields. This issue adds explicit `condition`/`outcome` dictionaries to every entry across all 9 documents that contains numeric thresholds, ranges, or named mechanical effects — enabling the Pinder.Rules engine (#446) to evaluate rules programmatically.
+
+This is Python-only work. No C# code changes. The output is 8 new `*-enriched.yaml` files plus the already-existing `rules-v3-enriched.yaml`.
+
+---
+
+## File Inventory
+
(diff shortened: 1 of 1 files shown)


decay256 (commented):
## Code Review

**Maturity level**: spec document (pre-implementation)
**Errors (must fix)**: 0
**Warnings (must fix)**: 0

| # | Location | Note |
|---|----------|------|
| 1 | `docs/specs/issue-444-spec.md:147-161` | Approach A is labeled "(preferred for tables)" but the paragraph immediately after says "The implementer should use **Approach B** (separate entries) as the primary approach for consistency". The contradiction is minor since the final instruction is clear (use B), but removing the "(preferred)" label from Approach A would reduce confusion for the implementer. |
| 2 | `docs/specs/issue-444-spec.md:480-481` | The "additive only" rule says no original field may be modified, but also says "type may be updated if the original type was incorrect." Example 3 shows `type` changing from `table` to `trap_activation`. This is a deliberate exception but the spec contradicts itself — consider making the exception explicit in the additive-only rule rather than as a parenthetical. |
| 3 | `docs/specs/issue-444-spec.md:106` | `callback_distance: 0` is used in Example 2 to mean "opener callback" but the vocabulary table defines `callback_distance` as "Turns since topic introduced" with example `4`. A value of 0 turns is ambiguous — could mean "this turn" rather than specifically "the opener." Consider a dedicated key like `is_opener: true` or document that 0 means opener. |

### Positive notes
- Extremely thorough spec with clear before/after examples for 4 different document types
- Controlled vocabulary tables for condition/outcome keys are well-defined with types and examples
- Edge cases section covers real scenarios (mixed tables, empty descriptions, duplicate IDs)
- Acceptance criteria are concrete and testable (AC3/AC4 accuracy check with 0 INACCURATE findings)
- Dependency section correctly identifies upstream (#443) and downstream (#445, #446) relationships

**VERDICT: APPROVE** (cannot formally approve own PR due to GitHub restriction — 0 errors, 0 warnings, approve intended)

PR #478, opened 2026-04-04
fix: InterestChangeBeat references conversation history (closes date-secured generic invite bug)

decay256:
The DateSecured narrative beat was generating generic invites like 'omg there's this new place from Instagram!' because the LLM had the opponent's character prompt but not the actual conversation history.

Now the beat prompt includes the last 6 messages from the conversation, so the LLM can reference specific details (shared jokes, topics discussed, running gags) when suggesting a meetup.

- InterestChangeContext: ConversationHistory + PlayerName fields added
- GameSession: passes history to context
- Adapter: passes history to builder
- Builder: injects last 6 messages as context block before the beat instruction
- Template: explicitly requires referencing something concrete from the conversation


decay256 (commented):
VERDICT: APPROVE — All 2752 tests pass. Five changes reviewed: (1) InterestChangeBeat now receives last 6 messages of conversation history via ConversationHistory/PlayerName on InterestChangeContext, threaded through GameSession→AnthropicAdapter→SessionDocumentBuilder. Clean. (2) Bio property added to CharacterProfile, parsed from character files, shown in playtest header. (3) PrintQuoted helper correctly prefixes every line with > for multi-line blockquotes. (4) SuccessDeliveryInstruction tightened — strong success constrained to phrasing/timing improvements only. (5) ASCII box UI replaced with clean one-block-per-option markdown format. No regressions.

PR #488, opened 2026-04-04
[Issue #444] Write tests: Rules DSL: enrich all 9 YAML files with explicit condition/outcome fields

### LINKED ISSUE #444: Rules DSL: enrich all 9 YAML files with explicit condition/outcome fields

decay256:
Fixes #444

## DoD Evidence
**Branch:** issue-444-write-tests-rules-dsl-enrich-all-9-yaml-
**Commit:** c4bdc47

## Test Summary
24 new spec compliance tests in `rules/tools/test_issue444_spec_compliance.py`:

- **stat_modifiers validation** (4 tests): Verifies `dict<string, int>` type, presence in items-pool and anatomy-parameters
- **accuracy_check.py execution** (2 tests): Runs script, verifies 0 INACCURATE findings
- **per-file enrichment minimums** (5 tests): Each major file has substantial enrichment counts
- **table-row splitting** (1 test): IDs follow §N.slug.qualifier convention
- **known rule completeness** (5 tests): Momentum, combos, tells, energy costs, horniness modifiers
- **snake_case convention** (2 tests): All condition/outcome keys follow naming convention
- **range ordering** (1 test): Range lower bounds ≤ upper bounds
- **empty dict detection** (2 tests): No empty condition/outcome dicts
- **total count thresholds** (2 tests): ≥150 enriched entries, ≥250 total entries

Supplements existing 35 tests in `test_issue444_enrichment.py` (total: 59 tests for issue #444).

### DIFF
--- a/rules/tools/test_issue444_spec_compliance.py
+++ b/rules/tools/test_issue444_spec_compliance.py
@@ -1,0 +1,501 @@
+#!/usr/bin/env python3
+"""
+Tests for Issue #444 — Spec compliance tests for Rules DSL enrichment.
+
+Written by test-engineer agent from docs/specs/issue-444-spec.md.
+Supplements test_issue444_enrichment.py with additional coverage for:
+- stat_modifiers dict<string, int> type validation (added post-initial tests)
+- accuracy_check.py execution (AC4)
+- per-file enrichment minimum thresholds
+- table-row splitting ID convention
+- enrichment completeness cross-checks
+
+Maturity: prototype — happy-path + key edge cases per acceptance criterion.
+"""
+
+import os
+import re
+import subprocess
+import sys
+from collections import Counter
+from typing import Any, Dict, List, Optional, Set
+
+import yaml
+
+# ---------------------------------------------------------------------------
+# Paths
+# ---------------------------------------------------------------------------
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+EXTRACTED_DIR = os.path.join(SCRIPT_DIR, '..', 'extracted')
+
+ENRICHED_FILES = [
+    'rules-v3-enriched.yaml',
+    'risk-reward-and-hidden-depth-enriched.yaml',
+    'async-time-enriched.yaml',
+    'traps-enriched.yaml',
+    'archetypes-enriched.yaml',
+    'character-construction-enriched.yaml',
+    'items-pool-enriched.yaml',
+    'anatomy-parameters-enriched.yaml',
+    'extensibility-enriched.yaml',
+]
+
+ORIGINAL_BASENAMES = [
+    'risk-reward-and-hidden-depth',
+    'async-time',
+    'traps',
+    'archetypes',
+    'character-construction',
+    'items-pool',
+    'anatomy-parameters',
+    'extensibility',
+]
+
+# Condition keys from spec vocabulary table
+CONDITION_KEYS_SPEC = {
+    'miss_range', 'beat_range', 'interest_range', 'need_range', 'level_range',
+    'timing_range', 'natural_roll', 'miss_minimum', 'streak', 'streak_minimum',
+    'action', 'stat', 'failed_stat', 'shadow', 'threshold', 'conversation_start',
+    'formula', 'shadow_points_per_penalty', 'dc', 'time_of_day', 'energy_below',
+    'trap_active', 'combo_sequence', 'callback_distance',
+}
+
+# Outcome keys from spec vocabulary table
+OUTCOME_KEYS_SPEC = {
+    'interest_delta', 'interest_bonus', 'roll_bonus', 'roll_modifier',
+    'dc_adjustment', 'xp_multiplier', 'xp_payout', 'risk_tier', 'tier',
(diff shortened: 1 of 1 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 2
**Warnings (must fix)**: 2

### Errors

| 1 | `rules/tools/test_issue444_spec_compliance.py:157,168` | AC4 tests silently pass when `accuracy_check.py` does not exist — `if not os.path.exists(script_path): return` makes the test a no-op, which means it claims to test AC4 but actually tests nothing if the file is missing. A test that silently passes is worse than no test. | Replace bare `return` with `pytest.skip("accuracy_check.py not found")` or better yet `assert os.path.exists(script_path), "AC4 requires accuracy_check.py to exist"` so the test fails loudly when the contract is violated. | Run tests with `accuracy_check.py` temporarily renamed — the test should skip or fail, not pass silently. |
| 2 | `rules/tools/test_issue444_spec_compliance.py:55-75` | `CONDITION_KEYS_SPEC` and `OUTCOME_KEYS_SPEC` are defined (copy-pasted from the spec vocabulary table) but **never referenced in any test assertion**. These are dead code. More importantly, this means there is **no test that validates condition/outcome keys are from the controlled vocabulary** — a critical spec requirement. The accuracy_check.py has its own (different, larger) key sets, but this test file never validates against them or the spec set. | Either (a) add a test that validates all condition/outcome keys in enriched files are members of `CONDITION_KEYS_SPEC ∪ {additional keys from accuracy_check.py}`, or (b) remove the dead constants and document that vocabulary validation is covered by accuracy_check.py (and the AC4 test covers it transitively). Option (a) is preferred. | `grep -n CONDITION_KEYS_SPEC rules/tools/test_issue444_spec_compliance.py` should show usage in a test method, not just the definition. |

### Warnings

| 1 | `rules/tools/test_issue444_spec_compliance.py:95` | `assert isinstance(v, (int, float))` for `stat_modifiers` values, but spec §Outcome Key Vocabulary explicitly says `dict<string, int>`. Accepting `float` makes the test weaker than the spec requires — it would pass if someone wrote `{"charm": 1.5}` which the spec does not allow. | Change to `assert isinstance(v, int)` to match the spec exactly. If YAML auto-parses some ints as floats, handle that edge case explicitly (e.g., check `v == int(v)`). | `grep "int, float" rules/tools/test_issue444_spec_compliance.py` should return no results. |
| 2 | `rules/tools/test_issue444_spec_compliance.py:42-50` | `ORIGINAL_BASENAMES` list is defined but never used anywhere in the file. Dead code in a test file signals incomplete implementation — was there a planned test that compares original vs enriched files? | Remove the dead constant, or add the test that was intended to use it (e.g., verifying each original file has a corresponding enriched file). | `grep -n ORIGINAL_BASENAMES rules/tools/test_issue444_spec_compliance.py` — should either not exist or be used in a test. |

| # | Location | Note |
|---|----------|------|
| 1 | `TestPerFileEnrichmentMinimums` | No minimum threshold tests for `archetypes-enriched.yaml` (41 enriched), `character-construction-enriched.yaml` (18 enriched), or `extensibility-enriched.yaml` (1 enriched). Consider adding for completeness in a future pass. |
| 2 | `test_accuracy_check_zero_inaccurate:185` | The final assertion `assert "PASS" in output or "pass" in output.lower() or result.returncode == 0` is redundant — `result.returncode == 0` was already asserted in the previous test, and this test already checked returncode would not be non-zero (it would fail at the subprocess call). The `or result.returncode == 0` clause makes this assertion always true if the script exits cleanly, even if it produces no PASS output and no INACCURATE count. Consider tightening. |

### Positive notes
- Good use of mutation-catching comments on each test — makes review intent clear.
- Error messages include file names and entry IDs — excellent for debugging failures.
- Range ordering validation (`TestRangeValueOrdering`) and empty dict checks (`TestEnrichedEntriesNotEmpty`) are good defensive tests.
- Snake_case validation for new keys is a smart spec-compliance check.
- Tests are behavior-focused and would survive refactoring of the enrichment implementation.

**VERDICT: REQUEST_CHANGES**

PR #504, opened 2026-04-04
[Issue #489] Write tests: Prompt: voice distinctness — explicit texting style constraint before option generation

### LINKED ISSUE #489: Prompt: voice distinctness — explicit texting style constraint before option generation
## Context
Character voices are bleeding together in option generation. Velvet is writing in Sable's register. The root fix is separating system prompts (#487), but even with that, the texting style needs to be more explicitly enforced.

## Problem
The TEXTING STYLE section is in a 3000-token system prompt. By the time the LLM reaches the task instruction, the style specifics are diluted. The LLM defaults to a warm generic Gen-Z texting voice.

## Fix
Extract the TEXTING STYLE block and inject it immediately before the task instruction in the user message:

```
CONVERSATION HISTORY
[...]

YOUR TEXTING STYLE — follow this exactly, no deviations:
{player_texting_style_fragment verbatim}

YOUR TASK
Generate 4 dialogue options in this exact voice.
```

This doubles the signal on texting style and puts it immediately adjacent to the task. The LLM can't skip it.

## Also: add a voice check instruction
```
Before writing each option, verify: does this sound exactly like the texting style above? If not, rewrite it.
```

## Acceptance Criteria
- [ ] `BuildDialogueOptionsPrompt` injects TEXTING STYLE verbatim before the task instruction
- [ ] `CharacterProfile` exposes `TextingStyleFragment` (parsed from the assembled prompt or design/examples/*.md)
- [ ] Run Velvet vs Sable: Velvet options use Velvet's register (lowercase-with-intent, precise, no 😭), Sable's response uses Sable's register (omg, 😭, fast-talk)
- [ ] Build clean

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Fixes #489

## DoD Evidence
**Branch:** issue-489-write-tests-prompt-voice-distinctness-ex
**Commit:** f913c8a

## Test Summary
23 spec-driven tests covering all acceptance criteria from `docs/specs/issue-489-spec.md`:

### AC2: CharacterProfile.TextingStyleFragment
- Returns provided value
- Defaults to empty string (not null)
- Null coalesced to empty
- Preserves exact value (emoji, unicode)

### AC2: DialogueContext.PlayerTextingStyle
- Returns provided value
- Defaults to empty string
- Null coalesced to empty

### AC1: SessionDocumentBuilder texting style injection
- Emits TEXTING STYLE block when non-empty
- Block appears before YOUR TASK
- Omits block when empty
- Injects style text verbatim
- Uses exact heading text
- Preserves unicode/emoji
- No truncation on long text
- Preserves pipe-joined fragments

### AC1: PromptTemplates voice check
- DialogueOptionsInstruction contains voice-check verification text

### Backward compatibility
- Default context (no style) works without TEXTING STYLE block
- YOUR TASK always present regardless of style

### Spec examples
- Example 1: Velvet with style
- Example 2: No style
- Example 3: Sable with emoji style

All tests include mutation comments documenting what specific code change each test would catch.

### DIFF
--- a/tests/Pinder.LlmAdapters.Tests/Issue489_VoiceDistinctnessSpecTests.cs
+++ b/tests/Pinder.LlmAdapters.Tests/Issue489_VoiceDistinctnessSpecTests.cs
@@ -1,0 +1,391 @@
+using System;
+using System.Collections.Generic;
+using Pinder.Core.Characters;
+using Pinder.Core.Conversation;
+using Pinder.Core.Stats;
+using Pinder.LlmAdapters;
+using Xunit;
+
+namespace Pinder.LlmAdapters.Tests
+{
+    /// <summary>
+    /// Spec-driven tests for Issue #489: voice distinctness — explicit texting style
+    /// constraint before option generation.
+    /// Tests are derived from docs/specs/issue-489-spec.md acceptance criteria and edge cases.
+    /// </summary>
+    public class Issue489_VoiceDistinctnessSpecTests
+    {
+        // ── Test-only helpers ──
+
+        private static StatBlock MakeStats()
+        {
+            return new StatBlock(
+                new Dictionary<StatType, int>
+                {
+                    { StatType.Charm, 5 },
+                    { StatType.Rizz, 3 },
+                    { StatType.Honesty, 2 },
+                    { StatType.Chaos, 1 },
+                    { StatType.Wit, 4 },
+                    { StatType.SelfAwareness, 0 }
+                },
+                new Dictionary<ShadowStatType, int>());
+        }
+
+        private static CharacterProfile MakeProfile(string textingStyleFragment = "")
+        {
+            return new CharacterProfile(
+                MakeStats(),
+                "system prompt",
+                "TestChar",
+                new TimingProfile(0, 1f, 0f, "neutral"),
+                1,
+                bio: "",
+                textingStyleFragment: textingStyleFragment);
+        }
+
+        private static DialogueContext MakeContext(
+            string playerTextingStyle = "",
+            string playerName = "Velvet",
+            string opponentName = "Sable",
+            int currentInterest = 10,
+            int currentTurn = 1)
+        {
+            return new DialogueContext(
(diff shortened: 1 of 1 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 0
**Warnings (must fix)**: 0

### Errors

None.

### Warnings

None.

| # | Location | Note |
|---|----------|------|
| 1 | `Issue489_VoiceDistinctnessSpecTests.cs:230-244` | `BuildDialogueOptionsPrompt_UsesPlayerStyle_NotOpponentStyle` — test name promises a negative assertion (opponent's style not used) but only asserts the player's style IS present. Consider adding a negative assertion for completeness, or renaming to `BuildDialogueOptionsPrompt_InjectsPlayerStyle`. |
| 2 | `Issue489_VoiceDistinctnessSpecTests.cs` | Whitespace-only style edge case missing — what happens with `"   "` (spaces only)? Should it be treated as empty and omit the block, or inject whitespace? Worth a future test. |
| 3 | `Issue489_VoiceDistinctnessSpecTests.cs:280,322` | Emoji preservation is tested twice (edge case test + Spec Example 3) with identical style strings. Minor duplication — both cover slightly different integration paths so acceptable, but could be consolidated. |

### Positive notes
- **Import vs. Rewrite: PASS** — all types (`CharacterProfile`, `DialogueContext`, `SessionDocumentBuilder`, `PromptTemplates`, `StatBlock`) are imported from real source modules. No local reimplementation of production logic.
- Every test has a mutation comment explaining what it would catch — good spec-driven discipline.
- Good edge case coverage: null coalescing, Unicode/emoji, long strings (1500 chars), pipe-joined fragments, backward compatibility.
- Ordering assertion (`TEXTING STYLE` before `YOUR TASK`) is a strong structural test that would survive most refactors.
- `[Theory]` with `[InlineData]` used appropriately for the always-contains-YOUR-TASK check.
- Tests are independent with no shared state.

**VERDICT: APPROVE**

PR #524, opened 2026-04-05
[Issue #516] Bug: CharacterLoader.ParseLevel reads wrong level — Velvet loads as 4 instead of 7

### LINKED ISSUE #516: Bug: CharacterLoader.ParseLevel reads wrong level — Velvet loads as 4 instead of 7
## Root cause
The level line in the prompt file is:
```
- Level: 7 (Veteran) | Level bonus: +3
```

`ParseLevel` extracts the number after `- Level: `. But the ARCHETYPES section also contains numbers (archetype counts like `×7`) and the STATS section contains numbers. If the level line appears AFTER some other numeric content, the parser might match the wrong line.

Or alternatively: the `CharacterAssembler` path (from `--player-def`) reads level from the JSON definition file, not from the prompt. The `CharacterLoader` path (from `--player name`) reads from the markdown. The test was `--player velvet` which uses `CharacterLoader`. 

Check what level value is in `design/examples/velvet-prompt.md` on the level line vs what the parser extracts:

```bash
grep 'Level:' /root/.openclaw/agents-extra/pinder/design/examples/velvet-prompt.md
```
→ Shows `- Level: 7 (Veteran) | Level bonus: +3`

But the session output showed Level 4 for Velvet. This means either:
1. There's a `Level:` line elsewhere in the file with a lower number that gets matched first, OR
2. The extraction is parsing '4' from something else

## Investigation
Run:
```bash
grep -n 'Level' /root/.openclaw/agents-extra/pinder/design/examples/velvet-prompt.md | head -10
```
to find all occurrences. The parser should match only the `- Level:` line (with leading dash).

## Acceptance Criteria
- [ ] Velvet loads as Level 7, not Level 4
- [ ] All 5 starters load correct levels
- [ ] Unit test for ParseLevel across all example files
- [ ] Build clean

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Fixes #516

## What was done

Updated stale level values in JSON character definition files (data/characters/*.json) that caused characters to load with incorrect levels. The JSON path (CharacterDefinitionLoader) is tried before the prompt file path (CharacterLoader), so the wrong JSON values were being used. Fixed: brick 6→9, velvet 4→7, zyx 4→1.

## How to test

```bash
dotnet test tests/Pinder.Core.Tests --filter CharacterLoaderTests
```

The new `ParseLevel_AllStarterCharacters_ReturnsCorrectLevel` theory test verifies all 5 characters against their prompt files.

## Changes

- `data/characters/brick.json`: level 6 → 9
- `data/characters/velvet.json`: level 4 → 7
- `data/characters/zyx.json`: level 4 → 1
- `tests/Pinder.Core.Tests/CharacterLoaderTests.cs`: added 5 ParseLevel tests (4 unit + 1 theory with 5 cases)

## Deviations from contract

None. The root cause was stale JSON data, not a code bug in ParseLevel itself.

### DIFF
--- a/data/characters/brick.json
+++ b/data/characters/brick.json
@@ -2,7 +2,7 @@
   "name": "Brick_haus",
   "gender_identity": "she/her",
   "bio": "CEO of knowing my worth. Don't waste my time.",
-  "level": 6,
+  "level": 9,
   "items": [
     "slicked-back-ponytail",
     "blazer-crop-top",
--- a/data/characters/velvet.json
+++ b/data/characters/velvet.json
@@ -2,7 +2,7 @@
   "name": "Velvet_Void",
   "gender_identity": "she/her",
   "bio": "if you have to ask, you're not invited.",
-  "level": 4,
+  "level": 7,
   "items": [
     "messy-bun-chopsticks",
     "oversized-band-tee",
--- a/data/characters/zyx.json
+++ b/data/characters/zyx.json
@@ -2,7 +2,7 @@
   "name": "Zyx_444",
   "gender_identity": "they/them",
   "bio": "the mushroom knows. you'll understand when you're ready.",
-  "level": 4,
+  "level": 1,
   "items": [
     "mushroom-cap-hat",
     "tie-dye-poncho",
--- a/tests/Pinder.Core.Tests/CharacterLoaderTests.cs
+++ b/tests/Pinder.Core.Tests/CharacterLoaderTests.cs
@@ -406,5 +406,74 @@ PERSONALITY
             string result = CharacterLoader.ParseBio(content);
             Assert.Equal("Scorpio. Living my best life. Send memes.", result);
         }
+
+        // ── ParseLevel tests ───────────────────────────────────────────
+
+        [Fact]
+        public void ParseLevel_StandardFormat_ReturnsLevel()
+        {
+            string content = @"```
+LEVEL
+- Level: 7 (Veteran) | Level bonus: +3
+
+EFFECTIVE STATS
+- Charm: +5
+```";
+            int result = CharacterLoader.ParseLevel(content);
+            Assert.Equal(7, result);
+        }
+
+        [Fact]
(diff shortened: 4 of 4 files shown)


decay256 (commented):
## Code Review

**Maturity level**: prototype
**Errors (must fix)**: 0
**Warnings (must fix)**: 2

### Warnings

| 1 | `tests/Pinder.Core.Tests/CharacterLoaderTests.cs:463-471` | `ParseLevel_AllStarterCharacters_ReturnsCorrectLevel` silently `return`s (passes green) when prompt files don't exist. In CI or any env without those files, all 5 theory cases pass without asserting anything — a test that tests nothing meaningful. | Replace `return;` with `Assert.Skip("Prompt files not available in this environment");` (xUnit v2.5+) so skipped cases are visible in test output, not falsely green. If xUnit version doesn't support Skip, use `[Fact(Skip = ...)]` or throw with a clear message. | Run tests without prompt files present — test should show as Skipped, not Passed. |
| 2 | `tests/Pinder.Core.Tests/CharacterLoaderTests.cs:464` | Hardcoded absolute path `/root/.openclaw/agents-extra/pinder/design/examples` couples the test to a specific machine layout. Won't work in CI or other dev environments. | Use a relative path from the repo root, an environment variable, or a test fixture that copies the files. At minimum mark this as an integration test category. | `grep -rn '/root/.openclaw' tests/` should return zero matches. |

| # | Location | Note |
|---|----------|------|
| 1 | `data/characters/zyx.json` | Zyx level changed 4→1 and brick 6→9 — neither mentioned in the issue title (which only cites Velvet). The theory test asserts these values so presumably correct, but worth a quick sanity check against the prompt files. |

### Positive notes
- The 4 unit tests for `ParseLevel` (standard, single-digit, double-digit, missing) are well-structured and cover the important parsing cases including the default-return path.
- Data fix is clean and minimal — correct root cause identified (stale JSON vs authoritative prompt files).
- Single focused commit with clear message.

**VERDICT: REQUEST_CHANGES**

PR #525, opened 2026-04-05
[Issue #517] Bug: ScoringPlayerAgent applies combo/tell bonus to options with near-0% success (inflates EV past trap cost)

### LINKED ISSUE #517: Bug: ScoringPlayerAgent applies combo/tell bonus to options with near-0% success (inflates EV past trap cost)
## Root cause
In session 011 turn 7, the SA option had:
- Base success probability: ~45% 
- But the scorer also applied Tell+2 and Recovery combo bonus
- Combined EV showed as 1.5 despite 0% showing in the table

The 0% was because the roll would nearly certainly TropeTrap (miss by 6-9 at DC16 with +2 mod = near-certain Spiral trap, 2-turn penalty). The combo bonus (+2 interest on recovery) was being applied to the expected success case even though the expected case was failure + trap activation.

## Fix
When computing expected cost of failure, if the miss margin would land in the TropeTrap range (6-9), apply the full trap cost:
- TropeTrap interest delta: −2
- Plus expected trap duration cost: ~1.5 turns of reduced effectiveness

The trap severity penalty in the EV formula must be applied BEFORE the combo/tell bonuses are added, not as an additive term.

Also: combo bonuses should not be applied when base success probability < 20%. A combo that fires 15% of the time is worth ≤ 0.15 × bonus, not the full combo bonus.

## Acceptance Criteria
- [ ] SA option with high trap risk correctly shows low/negative EV despite Tell+Combo bonuses
- [ ] Unit test: option with 15% success and TropeTrap-range miss scores lower than same option with 50% success
- [ ] Build clean

Depends on: #416 (ScoringPlayerAgent shadow growth risk)

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Fixes #517

## What was changed

**Root cause**: ScoringPlayerAgent used a flat DefaultFailCost=1.5 for all failure scenarios, regardless of miss margin distribution. Combined with combo bonuses at full value on near-impossible rolls, this inflated EV on terrible options.

**Fix**:
1. **Weighted failure cost** (ComputeWeightedFailCost): Maps each failing d20 result to its actual tier cost including trap activation (TropeTrap/Catastrophe/Legendary add 1.5 trap penalty).
2. **Combo scaling at low success**: When success < 20%, combo bonus scaled by successChance/0.20.

## How to test
```bash
dotnet test tests/Pinder.Core.Tests/ --filter ScoringPlayerAgent
```
88 tests pass (80 existing + 8 new). Full suite: 2979 passed, 17 skipped, 0 failed.

## Deviations from contract
None


decay256 (commented):
## Code Review

**Maturity level**: MVP
**Errors (must fix)**: 0
**Warnings (must fix)**: 0

| # | Location | Note |
|---|----------|------|
| 1 | `ScoringPlayerAgentEvTests.cs:280` | Test `LowSuccess_WithTellAndCombo_StillNegativeEv` asserts the plain option has negative EV but doesn't assert the bonused option is also negative — which is what the test name implies. Consider adding `Assert.True(decisionBonuses.Scores[0].ExpectedInterestGain < 0, ...)` to match the name. |
| 2 | `ScoringPlayerAgent.cs:369-373` | Fumble (miss 1-2) and Misfire (miss 3-5) both cost 1.0. Could be collapsed into a single `else if (missMargin < 6)` branch. Minor readability nit — the explicit tiers are arguably clearer for documentation purposes. |

### Positive notes
- `ComputeWeightedFailCost` math is correct: each failing roll is weighted equally (1/maxFailRoll), and when multiplied by `failChance` in the EV formula, produces the correct expected cost (totalCost/20).
- Combo scaling ramp (`successChance / 0.20`) is clean — linear degradation to 0 at 0% success, full value at 20%+.
- Tell bonus correctly flows through the roll modifier (`need`) rather than needing separate scaling — good design.
- Tests import from the real `Pinder.SessionRunner` module — no logic reimplementation.
- 8 tests cover the key behaviors: scaled combo at low success, unscaled combo at high success, weighted fail cost tiers, and integration scenarios comparing 15% vs 50% success.
- Good use of assertion messages with actual values for debugging.

**VERDICT: APPROVE**

decay256:

Issue #596, opened 2026-04-06
Vision: #572 Removing [RESPONSE] format tag risks conversational filler in output


decay256:
## Concern
Issue #572 proposes fixing the format tag leak by removing the `[RESPONSE]` wrapper requirement entirely (Option 2). However, without a structured format tag, LLMs frequently include conversational filler (e.g., "Here is what Velvet would say:") before the actual dialogue. The format tag exists to reliably extract the in-character dialogue from the LLM's meta-output.

## Why it matters
If the format tag is removed, we lose the ability to programmatically guarantee that only character dialogue is displayed. This violates the domain invariant that character dialogue must be clean and immersive.

## What needs to happen
- Do not remove the `[RESPONSE]` tag requirement.
- Instead, implement Option 1 (Fallback parsing): gracefully handle cases where the `[RESPONSE]` tag is present but the subsequent text is not wrapped in quotes.
- Enhance the parser to strip the `[RESPONSE]` marker and any leading/trailing whitespace if the regex fails.

## Acceptance Criteria
- [ ] Issue #572 is updated to pursue Option 1 (Fallback parsing) instead of Option 2.
- [ ] Parser strips `[RESPONSE]` marker when quotes are missing.

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

Issue #583, opened 2026-04-06
Vision: #536 Stateful LLM architecture violates Anthropic persona invariant and causes severe voice bleed


decay256:
## Concern
Issue #536 proposes accumulating all conversational turns (Options generation, Delivery generation, Opponent response) into a single Anthropic `messages[]` array. This violates the core Anthropic prompt engineering invariant: the `assistant` role must consistently represent the persona established by the system prompt. If the `assistant` role contains meta-JSON options, player delivery text, and opponent responses, the model's persona will fracture, causing catastrophic voice bleed (e.g., generating options in the opponent's voice).

## Why it matters
The domain invariant "Audio output must match the character's voice identity" (analogous to LLM persona consistency) will be broken. The model heavily anchors on its own past `assistant` messages. Mixing three different tasks/personas into one session thread destroys the isolation needed for high-quality character dialogue. The current architecture (stateless API calls that rebuild the prompt using the game's internal `_history`) is correct and already provides full context to the LLM.

## What needs to happen
- #536 must be closed or fundamentally rescoped to NOT use a persistent Anthropic `messages[]` array across different generation roles.
- If "in context" generation is desired, it must be achieved by enriching the stateless prompt (which already includes `_history`), not by stateful API accumulation.

## Acceptance Criteria
- [ ] Issue #536 is rescoped to avoid stateful `messages[]` accumulation across different personas.
- [ ] Alternatively, #536 is closed as architecturally flawed.

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

Issue #579, opened 2026-04-05
Vision: #574 CharacterDefinitionLoader skips bio param — CharacterProfile.Bio always empty via assembler path


decay256:
## Concern
`CharacterDefinitionLoader.Parse()` reads the bio from JSON (line 69) and passes it to `PromptBuilder.BuildSystemPrompt()` (line 92), but does NOT pass it to the `CharacterProfile` constructor (line 101):

```csharp
return new CharacterProfile(
    fragments.Stats, systemPrompt, name, fragments.Timing, level,
    textingStyleFragment: textingStyle);  // bio parameter skipped!
```

The `CharacterProfile` constructor has `bio = ""` as a default, so the assembler pipeline always produces profiles with empty bios. This is the real root cause of the empty bio display — NOT `CharacterLoader.ParseBio` (which was already fixed in #513).

Since the session runner tries `CharacterDefinitionLoader` first (line 117 of Program.cs) and only falls back to `CharacterLoader` on failure, most characters load through the assembler path and have empty bios.

## Why it matters
- Playtest headers show empty Bio rows, making character identity opaque
- The bio IS loaded from JSON — it's just dropped on the floor before reaching CharacterProfile
- Issue #574's AC says "verify fix from sprint landed" which may lead the implementer to look at the wrong code path

## What needs to happen
`CharacterDefinitionLoader` line 101 must pass the `bio` variable to the `CharacterProfile` constructor:
```csharp
return new CharacterProfile(
    fragments.Stats, systemPrompt, name, fragments.Timing, level,
    bio: bio,
    textingStyleFragment: textingStyle);
```

## Acceptance Criteria
- [ ] `CharacterDefinitionLoader.Parse()` passes the bio value to `CharacterProfile` constructor
- [ ] Characters loaded via the assembler path (`--player gerald`) display their bio in playtest output
- [ ] Unit test verifies `CharacterDefinitionLoader.Parse()` produces a profile with non-empty `Bio`

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

Issue #576, opened 2026-04-05
Vision: #575 GameSession builds system prompt by string concatenation — SessionSystemPromptBuilder.Build() never called


decay256:
## Concern
GameSession line 174 builds the stateful session system prompt as:
```csharp
var systemPrompt = _player.AssembledSystemPrompt + "\n\n---\n\n" + _opponent.AssembledSystemPrompt;
```

This bypasses `SessionSystemPromptBuilder.Build()` which was created in #543 specifically to inject game vision, world rules, player/opponent roles, meta contract, and writing rules. The `GameDefinition.PinderDefaults` content (dating frame, romantic pursuit direction, resistance rules) never reaches the LLM.

This is the **root cause** of #575 (no dating frame) and likely contributes to #531 (no dating dynamic).

## Why it matters
Without the game definition in the system prompt, the LLM has no concept that:
- This is a dating app conversation
- The player is pursuing, the opponent is evaluating
- Characters are real people who don't know about game mechanics
- Resistance should scale with interest level

The entire Session Architecture sprint (#541-#544) built infrastructure for this, but the final wiring step was missed.

## What needs to happen
`GameSession` must call `SessionSystemPromptBuilder.Build(playerPrompt, opponentPrompt, gameDef)` when starting a stateful conversation. This requires either:
1. Adding `GameDefinition?` to `GameSessionConfig` (preferred — keeps Core zero-dep by accepting the built string)
2. Or having the caller (session-runner) build the prompt and pass it via a new config field

## Acceptance Criteria
- [ ] Stateful session system prompt includes all 5 sections from `SessionSystemPromptBuilder.Build()`: game vision, world rules, player character, opponent character, meta contract + writing rules
- [ ] `GameDefinition.PinderDefaults` content (or loaded YAML) is present in the prompt sent to Anthropic
- [ ] Unit test verifies the system prompt contains game vision text when stateful adapter is used

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

PR #598, opened 2026-04-06
[Issue #572] Vision Review (attempt 2): Sprint 10

decay256:
Refs #572

## DoD Evidence
**Branch:** issue-572-vision-review-attempt-1-sprint-10
**Commit:** d20afff

# CPO Strategic Review — Sprint 10

## Alignment: ⚠️ ADVISORY
This sprint is focused on critical infrastructure and UI bugs that affect the product vision. However, there are significant architectural and scoping concerns. Issue #536 proposes stateful generation which will corrupt persona integrity, violating the core design of Anthropic's message context. Removing the format tag in #572 exposes the user to raw LLM meta-talk. Furthermore, #574 misidentifies the root cause of empty bios, and #575 needs a robust architectural fix rather than string concatenation. I have filed 4 vision concern issues to address these risks.

### Stateful Session Architecture (#536)
- User opens session → GameSession maintains a single Anthropic `messages[]` array → Options generated → Player message added → Opponent response added
- Required fields: System prompt, `user` messages, `assistant` messages
- ⚠️ Missing constraint: The `assistant` role is supposed to represent a single persona. Mixing JSON options, player text, and opponent text into one stateful `messages[]` array breaks the LLM's persona consistency.

### Game Definition Injection (#575)
- Host loads `game-definition.yaml` → `SessionSystemPromptBuilder.Build()` is supposed to be called → System prompt injected into LLM
- Required fields: Game vision, world rules, character profiles
- ⚠️ Missing: `GameSession` bypasses the prompt builder, so the game definition fields never flow into the LLM system prompt.

### Bio Loading (#574)
- JSON definition read → `CharacterDefinitionLoader` parses bio → `CharacterProfile` constructor called
- Required fields: Stats, Level, Bio, TextingStyle
- ⚠️ Missing: Bio is parsed but dropped before being passed to `CharacterProfile`, resulting in an empty bio at runtime.

## Unstated Requirements
- Removing the `[RESPONSE]` wrapper (proposed in #572) will eventually lead to the user seeing raw LLM conversational filler, which breaks the immersive dating app UI. The user expects pure in-character dialogue.
- When NarrativeBeat is removed (#573), the user still needs some indication of a vibe shift in the UI, even if it's not LLM-generated text.

## Domain Invariants
- **LLM Persona Integrity**: The Anthropic `assistant` role must exclusively represent the character voice established by the system prompt.
- **Clean Dialogue**: Game UI must display raw character dialogue, stripped of any LLM formatting or conversational preamble.

## Gaps
- **Assumption**: Issue #536 assumes that stateless context generation is flawed, but the stateless design already injects the full `_history`, and stateful generation will cause voice bleed.
- **Missing**: Issue #574 fixes a fallback loader but misses the primary assembler path, which is dropping the bio property.
- **Unnecessary**: Removing the formatting tag entirely in #572 is an overcorrection that introduces new risks.

## Recommendations
1. Rescope #536 to retain stateless generation, utilizing `_history` to provide context while preserving persona integrity (see #583).
2. For #572, do not remove the tag; instead implement a fallback parser that handles missing quotes (see #596).
3. Correct #574 to fix `CharacterDefinitionLoader` so it passes the bio to the profile (see #579).
4. For #575, wire `SessionSystemPromptBuilder.Build()` into `GameSession` properly instead of raw string concatenation (see #576).

**Verdict:** ADVISORY
**Concerns raised:** #583, #596, #579, #576

### DIFF
--- a/docs/sprint-10-cpo-review.md
+++ b/docs/sprint-10-cpo-review.md
@@ -1,0 +1,39 @@
+# CPO Strategic Review — Sprint 10
+
+## Alignment: ⚠️ ADVISORY
+This sprint is focused on critical infrastructure and UI bugs that affect the product vision. However, there are significant architectural and scoping concerns. Issue #536 proposes stateful generation which will corrupt persona integrity, violating the core design of Anthropic's message context. Removing the format tag in #572 exposes the user to raw LLM meta-talk. Furthermore, #574 misidentifies the root cause of empty bios, and #575 needs a robust architectural fix rather than string concatenation. I have filed 4 vision concern issues to address these risks.
+
+## Data Flow Traces
+### Stateful Session Architecture (#536)
+- User opens session → GameSession maintains a single Anthropic `messages[]` array → Options generated → Player message added → Opponent response added
+- Required fields: System prompt, `user` messages, `assistant` messages
+- ⚠️ Missing constraint: The `assistant` role is supposed to represent a single persona. Mixing JSON options, player text, and opponent text into one stateful `messages[]` array breaks the LLM's persona consistency.
+
+### Game Definition Injection (#575)
+- Host loads `game-definition.yaml` → `SessionSystemPromptBuilder.Build()` is supposed to be called → System prompt injected into LLM
+- Required fields: Game vision, world rules, character profiles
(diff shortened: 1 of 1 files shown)


decay256:

PR #756, opened 2026-04-24
feat(session-setup): extract matchup/stake helpers + add per-stage turn hooks

decay256:
Track B dependency for pinder-web#114. Adds Pinder.SessionSetup library (MatchupAnalyzer + stake generator) and IProgress<TurnProgressEvent> hooks in GameSession.ResolveTurnAsync so the web backend can emit per-stage SSE events. session-runner migrated to consume the new library.


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

### Blocking
(none)

### Non-blocking

- **[src/Pinder.SessionSetup/LlmMatchupAnalyzer.cs:67-73, 101-105]** The bare `catch` on cache read failure and the bare `catch` on the LLM transport call discard the exception entirely. For an MVP with real paid LLMs, a one-line `// TODO` or a minimal `ILogger` parameter would let us see why a matchup silently degrades to `null` (transport failure? bad cache file? rate limit?). Behavioural parity with the legacy helper is defensible, but parity is the current state of the problem, not a fix.

- **[src/Pinder.SessionSetup/LlmMatchupAnalyzer.cs:41-48 + LlmStakeGenerator.cs:28-38]** Both constructors accept `CancellationToken cancellationToken = default` on the public method, but the underlying `ILlmTransport.SendAsync(systemPrompt, userPrompt, temperature, maxTokens)` signature (as used here) has no CT parameter, so the token passed in is effectively ignored for the actual LLM HTTP call. This looks cancellable from the outside but isn't. Either drop the parameter (matches reality) or thread the CT through `ILlmTransport`.

### Positive notes

- Clean extraction: the new `Pinder.SessionSetup` library isolates matchup + stake generation behind `IMatchupAnalyzer` / `IStakeGenerator` with a provider-agnostic `ILlmTransport` seam. `session-runner/Program.cs` is migrated onto the new library — no copy-paste remains.
- `TurnProgressStage` / `TurnProgressEvent` are a tight, dependency-free way to expose the multi-LLM pipeline stages. The enum names are 1:1 with the SSE event names on the web side, which keeps the contract auditable. XML doc on each stage is welcome.
- The non-progress `ResolveTurnAsync(int)` overload is preserved and simply forwards to `ResolveTurnAsync(int, IProgress<…>?)` — no breaking change for existing callers (including `Pinder.Core.Tests` and `session-runner`).
- New `GameSession` properties (`TurnNumber`, `IsEnded`, `Outcome`, `ConversationHistory`) expose previously-private state cleanly. The `ConversationHistory` doc comment pins the invariant that the list is only appended inside `ResolveTurnAsync`, which the web-side `/log` endpoint relies on.
- `NullTrapRegistry` promoted to `Pinder.Core.Traps` with XML doc — one less piece of session-runner private wiring that the API had to duplicate.
- `LlmMatchupAnalyzer.Options` / `LlmStakeGenerator.Options` with sensible defaults (temperature, max tokens, cache dir) is the right surface for an MVP — tunable but not required.
- File cache write failure is swallowed non-fatally, and a corrupt cache read falls through to regeneration. Both are the right calls.
- `CharacterDefinitionLoader` move to `Pinder.SessionSetup` (with the tests updated to also reference the new project) is a namespace rename only; no behavioural drift visible in the diff.
- Test projects reference `Pinder.SessionSetup` correctly; existing loader tests continue to cover the moved code (they exercise `CharacterDefinitionLoader.Load(...)` as before).

**Verdict**: APPROVE

Issue #144, opened 2026-03-28
Arch concern: Energy system has infrastructure but no defined consumer — what costs energy?


decay256:
## Problem

The Sprint 5 architecture defines energy infrastructure:
- `IGameClock.RemainingEnergy` (property)
- `IGameClock.ConsumeEnergy(int amount)` (method)
- `IGameClock.ReplenishAtMidnight()` (daily reset)
- `ConversationRegistry.ConsumeEnergy()` (delegation)

But **no contract specifies what costs energy**. The §async-time rules say '15-20 energy per day' but the contracts don't define:
- Does each turn cost 1 energy?
- Does starting a conversation cost energy?
- What happens when energy reaches 0? (Can the player still act?)

## Impact

Implementers will build the energy tracking system but nothing will call `ConsumeEnergy()`. The system will be dead code until someone defines the consumption rules.

## Recommendation

For prototype: add a simple rule like 'each Speak/Read/Recover turn costs 1 energy, Wait is free, 0 energy = forced Wait' to the GameSession or ConversationRegistry contract. This makes the energy system testable end-to-end.

decay256:
Context update 2026-04-11: Energy budget is a multiplayer/async feature per rules §12 (ConversationRegistry, 15-20 turns/day, cross-chat shadow bleed). In single-session mode it has no defined consumer and that is correct — energy is scaffolding for a future feature. No action needed until multiplayer is in scope.

decay256:
Closing as **deferred — out of scope for the prototype**, not abandoned.

Rationale and notes on what a future version will need to decide first are recorded in [`docs/future-scope.md`](https://github.com/decay256/pinder-core/blob/main/docs/future-scope.md) (commit 7a9934e).

Re-open or file a fresh issue when this becomes in-scope.

PR #761, opened 2026-04-25
feat(setup-stream): AnthropicStreamingTransport — Anthropic Messages SSE (#149)

decay256:
Implements #149 (filed against `decay256/pinder-web`; code change lives here in `pinder-core`).

Adds `Pinder.LlmAdapters.Anthropic.AnthropicStreamingTransport`, the concrete `IStreamingLlmTransport` (added in #144) implementation backed by the Anthropic Messages API with `stream: true`.

## What was implemented

- New class `AnthropicStreamingTransport : IStreamingLlmTransport, IDisposable` (mirrors `AnthropicTransport` constructor pattern: `(apiKey, model)` with internally-owned `HttpClient`, plus `(apiKey, model, httpClient)` for tests).
- Builds the same `MessagesRequest` as the non-streaming sibling via `AnthropicRequestBuilders.BuildMessagesRequest`, then injects `"stream": true` into the JSON payload.
- Issues the request with `HttpCompletionOption.ResponseHeadersRead` so the body can be consumed as it arrives.
- SSE parser:
  - Reads line-by-line via `StreamReader.ReadLineAsync()`. Accumulates `data:` lines until a blank-line frame boundary, then dispatches.
  - For `data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"..."}}`, yields `delta.text` (in arrival order).
  - All other event types (`message_start`, `content_block_start`, `ping`, `content_block_stop`, `message_delta`, `message_stop`, and anything unknown) are silently ignored — never throws on them.
  - Stream terminates on EOF; no `[DONE]` sentinel (Anthropic doesn't send one).
- Error mapping:
  - HTTP open-time non-2xx → `LlmTransportException` with status code + truncated body (≤1KB).
  - Mid-stream `event: error` frame → `LlmTransportException` with `"<type>: <message>"` from `data.error`.
  - Network/IO mid-stream → `LlmTransportException` wrapping the inner exception.
  - `OperationCanceledException` propagates unchanged (cancellation is not a transport failure).
- Cancellation: a `CancellationTokenRegistration` disposes the underlying response stream when the token fires, since `ReadLineAsync` on `netstandard2.0` does not accept a `CancellationToken`. Pre-cancellation is checked deterministically before the `StreamReader` is constructed.

## How to test it

Local unit tests use a fake `HttpMessageHandler` and a `ChunkedStream` that delivers canned SSE bytes across multiple read calls — so the parser is exercised against realistic frame-split scenarios.

```
dotnet test tests/Pinder.LlmAdapters.Tests/Pinder.LlmAdapters.Tests.csproj \
  --filter "FullyQualifiedName~AnthropicStreamingTransport"
```

Result: **17 passed / 0 failed** (full output in DoD evidence below).

## Coverage

| Test | What it asserts |
|------|-----------------|
| `YieldsTextDeltasInOrder` | Multi-fragment stream reassembles in arrival order. |
| `RequestBodyContainsStreamTrue` | Outbound JSON has `"stream":true`, plus model / max_tokens / temperature wired correctly. |
| `FrameSplitAcrossChunks_StillReassembles` | Single SSE frame split into 3 byte-chunks across separate reads still yields the full text. |
| `UnknownEventTypes_AreIgnored` | `ping` / `message_start` / `weird_future_event` / `message_delta` / `message_stop` interleaved with one `text_delta` → exactly 1 yielded fragment, no throw. |
| `NonTextDelta_IsIgnored` | `content_block_delta` with `delta.type == "input_json_delta"` is ignored (tool-call deltas don't leak into text stream). |
| `Http401/500/400` | All open-time non-2xx codes map to `LlmTransportException`, message includes status + body excerpt. |
| `Http400_BodyTruncatedTo1KB` | 4KB body is truncated, exception message includes `"truncated"` marker and isn't itself 4KB+. |
| `MidStreamErrorEvent_ThrowsLlmTransportException` | Provider error frame mid-stream throws; partial fragments yielded *before* the error are preserved. |
| `CancellationBetweenFrames_ClosesCleanly` | Cancelling between fragments throws `OperationCanceledException` and disposes the underlying response stream. |
| `PreCancelledToken_ThrowsOperationCanceledNotTransportException` | Pre-cancelled token throws OCE — not wrapped as a transport failure. |
| Constructor null/whitespace guards | Match `AnthropicTransport` semantics. |
| `Dispose` idempotency / borrowed-client behaviour | Owned `HttpClient` disposed once; borrowed not disposed. |

## Sample SSE log

A captured-style sample is included in the class-level xmldoc on `AnthropicStreamingTransport` (synthesized from the spec — no live API call). It shows the canonical successful sequence (`message_start` → `content_block_start` → `ping` → `content_block_delta`×N → `content_block_stop` → `message_delta` → `message_stop`) plus the `error` frame shape.

## Out of scope (per #149)

- Wiring into `LlmProviderFactory` — separate issue.
- Other providers (OpenAI streaming etc.) — separate issue.

## Deviations from contract

None. `IStreamingLlmTransport.SendStreamAsync(systemPrompt, userMessage, temperature=0.9, maxTokens=1024, ct)` implemented exactly as defined in #144.

## Notes for the orchestrator

Sub-module bump in `pinder-web` is **not** done here — the orchestrator handles that after merge.

Refs decay256/pinder-web#149.

### DIFF
--- a/agent.log
+++ b/agent.log
@@ -1090,3 +1090,4 @@
 {"ts":"2026-04-12T20:22:50+00:00","agent":"backend-engineer","issue":"857","component":"","action":"completed","correlationId":"decompose-src-Pinder.LlmAdapters-Anthropic-AnthropicResponseParsers.cs-1776024849331","detail":"PR opened — correlationId: decompose-src-Pinder.LlmAdapters-Anthropic-AnthropicResponseParsers.cs-1776024849331","commit":"9fbacb4"}
 {"ts":"2026-04-25T07:10:27+00:00","agent":"backend-engineer","issue":"144","action":"started","detail":"Starting: streaming overloads on IMatchupAnalyzer + IStakeGenerator","commit":null}
 {"ts":"2026-04-25T07:20:01+00:00","agent":"backend-engineer","issue":"144","action":"completed","detail":"PR #759 opened — streaming overloads on IMatchupAnalyzer + IStakeGenerator","commit":"8b6bd84","prUrl":"https://github.com/decay256/pinder-core/pull/759"}
+{"ts":"2026-04-25T17:02:24Z","agent":"backend-engineer","issue":"149","action":"started","detail":"AnthropicStreamingTransport — Anthropic Messages SSE","commit":null}
--- a/agent.log
+++ b/agent.log
@@ -1091,3 +1091,4 @@
 {"ts":"2026-04-25T07:10:27+00:00","agent":"backend-engineer","issue":"144","action":"started","detail":"Starting: streaming overloads on IMatchupAnalyzer + IStakeGenerator","commit":null}
 {"ts":"2026-04-25T07:20:01+00:00","agent":"backend-engineer","issue":"144","action":"completed","detail":"PR #759 opened — streaming overloads on IMatchupAnalyzer + IStakeGenerator","commit":"8b6bd84","prUrl":"https://github.com/decay256/pinder-core/pull/759"}
 {"ts":"2026-04-25T17:02:24Z","agent":"backend-engineer","issue":"149","action":"started","detail":"AnthropicStreamingTransport — Anthropic Messages SSE","commit":null}
+{"ts":"2026-04-25T17:09:43Z","agent":"backend-engineer","issue":"149","action":"completed","detail":"PR #761 opened — AnthropicStreamingTransport (Anthropic Messages SSE)","commit":"e63b3eb","prUrl":"https://github.com/decay256/pinder-core/pull/761"}
--- a/src/Pinder.LlmAdapters/Anthropic/AnthropicStreamingTransport.cs
+++ b/src/Pinder.LlmAdapters/Anthropic/AnthropicStreamingTransport.cs
@@ -1,0 +1,416 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Pinder.Core.Interfaces;
+using Pinder.LlmAdapters.Anthropic.Dto;
+
+namespace Pinder.LlmAdapters.Anthropic
+{
+    /// <summary>
+    /// Streaming counterpart to <see cref="AnthropicTransport"/>. Implements
+    /// <see cref="IStreamingLlmTransport"/> on top of the Anthropic Messages
+    /// API with <c>stream: true</c>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// <b>Wire format.</b> The Anthropic Messages API streams responses as
+    /// Server-Sent Events. A typical successful stream looks like:
+    /// </para>
+    /// <code>
+    /// event: message_start
+    /// data: {"type":"message_start","message":{"id":"msg_01","model":"claude-sonnet-4-...","role":"assistant","content":[],"usage":{"input_tokens":25,"output_tokens":1}}}
+    ///
+    /// event: content_block_start
+    /// data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}
+    ///
+    /// event: ping
+    /// data: {"type":"ping"}
+    ///
+    /// event: content_block_delta
+    /// data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello"}}
+    ///
+    /// event: content_block_delta
+    /// data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" world"}}
+    ///
+    /// event: content_block_stop
+    /// data: {"type":"content_block_stop","index":0}
+    ///
+    /// event: message_delta
+    /// data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":15}}
+    ///
+    /// event: message_stop
+    /// data: {"type":"message_stop"}
+    /// </code>
+    /// <para>
+    /// On failure the server may emit:
+    /// </para>
+    /// <code>
+    /// event: error
+    /// data: {"type":"error","error":{"type":"overloaded_error","message":"Overloaded"}}
+    /// </code>
+    /// <para>
+    /// <b>Contract.</b>
+    /// </para>
+    /// <list type="bullet">
+    ///   <item>Yields only <c>content_block_delta</c> frames whose
+    ///   <c>delta.type == "text_delta"</c> — the <c>delta.text</c> field, in
+    ///   arrival order. Multi-chunk text reassembles by concatenation at the
+    ///   call-site.</item>
+    ///   <item>All other event types (<c>message_start</c>,
+    ///   <c>content_block_start</c>, <c>ping</c>, <c>content_block_stop</c>,
+    ///   <c>message_delta</c>, <c>message_stop</c>, unknown) are silently
+    ///   ignored. The stream terminates when the underlying response stream
+    ///   ends; no sentinel is required (Anthropic does not send <c>[DONE]</c>).</item>
+    ///   <item>An open-time non-2xx HTTP response throws
+    ///   <see cref="LlmTransportException"/> with the status code and a
+    ///   truncated body excerpt (≤1KB).</item>
+    ///   <item>A mid-stream <c>event: error</c> frame throws
+    ///   <see cref="LlmTransportException"/> with the provider error message.</item>
+    ///   <item>Mid-stream network/IO failures are wrapped in
+    ///   <see cref="LlmTransportException"/> preserving the inner exception.</item>
+    ///   <item>The supplied <see cref="CancellationToken"/> is propagated to
+    ///   the HTTP call and observed between SSE frames; cancellation closes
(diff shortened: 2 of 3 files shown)


decay256 (commented):
## QA Verdict: PASS

**Issue:** decay256/pinder-web#149 — `[setup-stream] AnthropicStreamingTransport — Anthropic Messages SSE`
**PR:** decay256/pinder-core#761
**Reviewer:** QA Engineer agent
**Maturity:** prototype

### Test Results

| Suite | Outcome |
|---|---|
| `Pinder.LlmAdapters.Tests` (filter `~AnthropicStreamingTransport`) | **17 passed / 0 failed / 0 skipped** (192 ms) |
| `Pinder.Core.Tests` (full) | **2402 passed / 0 failed / 18 skipped** (6 s) — skips are pre-existing rules-spec items, unrelated to #149 |

### Acceptance Criteria → Test Mapping

| # | Acceptance Criterion | Test / Evidence | Status |
|---|---|---|---|
| 1 | `AnthropicStreamingTransport : IStreamingLlmTransport, IDisposable` | `src/Pinder.LlmAdapters/Anthropic/AnthropicStreamingTransport.cs:84` — `public sealed class AnthropicStreamingTransport : IStreamingLlmTransport, IDisposable`; `Dispose_OwnedHttpClient_DoesNotThrow`, `Dispose_BorrowedHttpClient_DoesNotDisposeIt` | ✅ |
| 2 | Constructor mirror of `AnthropicTransport` (apiKey + model, optional injected `HttpClient`) | Two ctors at lines 100 / 115; signatures match `AnthropicTransport` lines 20 / 29. Covered by `Constructor_NullApiKey_Throws`, `Constructor_WhitespaceApiKey_Throws`, `Constructor_NullModel_Throws`, `Constructor_NullHttpClient_Throws` | ✅ |
| 3 | Calls Messages with `stream: true`; yields `delta.text` for each `content_block_delta` w/ `text_delta` | `SendStreamAsync_RequestBodyContainsStreamTrue`, `SendStreamAsync_YieldsTextDeltasInOrder`, `SendStreamAsync_FrameSplitAcrossChunks_StillReassembles` | ✅ |
| 4 | Ignore unknown event types (`ping`, `message_start`, `message_stop`, …) — never throw | `SendStreamAsync_UnknownEventTypes_AreIgnored`, `SendStreamAsync_NonTextDelta_IsIgnored` | ✅ |
| 5 | Honour `CancellationToken`; cancellation closes underlying response | `SendStreamAsync_CancellationBetweenFrames_ClosesCleanly`, `SendStreamAsync_PreCancelledToken_ThrowsOperationCanceledNotTransportException` | ✅ |
| 6a | HTTP open-time non-2xx → `LlmTransportException` w/ status code + truncated body | `SendStreamAsync_Http401_Throws_LlmTransportException_WithStatus`, `SendStreamAsync_Http500_Throws_LlmTransportException_WithStatus`, `SendStreamAsync_Http400_BodyTruncatedTo1KB` | ✅ |
| 6b | Mid-stream `event: error` frame → `LlmTransportException` | `SendStreamAsync_MidStreamErrorEvent_ThrowsLlmTransportException` | ✅ |
| 6c | Network/IO mid-stream → `LlmTransportException` wrapping inner | Implementation present (xmldoc lines 76‑78, code path wraps `IOException`) but not exercised by a dedicated unit test. See note below. | ⚠️ minor gap |
| 7 | Unit tests with fake `HttpMessageHandler` returning canned SSE bytes (tokens-in-order, cancellation, HTTP error, mid-stream error, unknown event types) | All five categories are covered; see rows 3–6b. The fake handler streams chunks via a `ChunkedStream` so the parser sees realistic partial reads — including a frame split across 3 byte boundaries (`SendStreamAsync_FrameSplitAcrossChunks_StillReassembles`). | ✅ |
| 8 | xmldoc on the class enumerates the contract | `AnthropicStreamingTransport.cs:16‑83` — full xmldoc block: wire format example, contract bullets (text-only yields, ignored event types, open-time errors, mid-stream errors, IO wrapping, cancellation semantics) | ✅ |

### Notes / Minor Gaps

- **AC 6c (mid-stream IOException wrapping)** is documented in the class xmldoc and implemented in the read loop but has no dedicated test. The `ChunkedStream` test fake never throws mid-read, so the IO-wrapping branch is uncovered. At **prototype** maturity this is acceptable — the happy / cancellation / HTTP-error / provider-error branches are all tested and the IO branch is straightforward — but it's the one obvious add for an MVP-level uplift.
- All other AC items are covered by deterministic tests. No flakes observed.
- Out-of-scope items (factory wiring, other providers) correctly absent from this PR.

### Verdict

**PASS** — every AC bullet is implemented and (with one minor exception called out above) covered by a passing unit test. The one uncovered branch is the IOException-wrap path; the AC requires the *behaviour* to exist (it does, by inspection of the xmldoc + code) but the test plan in the AC does not explicitly enumerate it, so I'm not blocking on it at prototype maturity.

Recommend merge after Code Review and Security Review sign off.

decay256 (commented):
## Code Review

**Maturity level**: prototype
**Blocking issues**: 0
**Non-blocking issues**: 6

### Blocking
*(none)*

### Non-blocking

- **[AnthropicStreamingTransport.cs:204–230] Dispose-on-cancel wake-up is not actually exercised by tests.** `CancellationBetweenFrames_ClosesCleanly` cancels between frames; the next loop iteration's `cancellationToken.ThrowIfCancellationRequested()` at line 218 catches it BEFORE `ReadLineAsync` is called. The harder path — cancellation firing mid-`ReadLineAsync` and the dispose registration wake-up causing the read to throw `ObjectDisposedException`/`IOException` — is the whole reason the workaround exists, and it's untested. A test using a `ChunkedStream` with a long inter-chunk delay (>200ms) and cancellation triggered after the StreamReader has clearly entered the blocking read would exercise it. Until that test exists, the catch arms at lines 234–245 are speculative.

- **[AnthropicStreamingTransport.cs:381–387] `ConfigureHeaders` silently drops the new api key on a borrowed HttpClient that already has `x-api-key` set.** The `Contains` check makes header registration idempotent, but if a caller reuses a single `HttpClient` across two transports with different API keys, the second key is silently ignored. For the test-only secondary ctor this is mostly fine, but worth a one-line xmldoc note: "if a different api-key is already present on a shared client, the existing one wins."

- **[AnthropicStreamingTransport.cs:208–212] Stream is double-disposed.** `using (stream)` on the outside and `leaveOpen: false` on the StreamReader (line 213) both dispose the stream. Stream.Dispose is documented as idempotent so this is harmless, but it's confusing — pick one (e.g. `leaveOpen: true` and let the outer `using` own it).

- **[AnthropicStreamingTransport.cs:152] Mid-stream network/IO failure → LlmTransportException is implemented but untested.** The acceptance criteria call out "Network/IO mid-stream → throw `LlmTransportException` wrapping inner". The catch block at lines 240–246 handles it but no canned test forces a mid-stream IO error (e.g. a `ChunkedStream` that throws `IOException` after delivering one frame). Easy to add.

- **[AnthropicStreamingTransport.cs:155] `JObject.FromObject(request)` + `payload["stream"] = true` round-trips the whole request through reflection on every call.** A `[JsonProperty("stream")]` field on `MessagesRequest` (or a streaming-specific request DTO) would be cleaner. Acceptable for prototype, but flag for cleanup before product maturity.

- **[AnthropicStreamingTransportTests.cs:391–396] `Dispose_BorrowedHttpClient_DoesNotDisposeIt` only asserts the variable is non-null.** It doesn't actually verify the borrowed HttpClient is still usable post-Dispose (e.g. by sending a second request through it). The assertion `Assert.NotNull(http)` is a tautology — `http` is a local non-null variable.

### Positive notes

- All Anthropic SSE event types from the spec (`message_start`, `content_block_start`, `ping`, `content_block_delta`, `content_block_stop`, `message_delta`, `message_stop`, `error`) handled with explicit ignore-by-default for unknown future events. Test `UnknownEventTypes_AreIgnored` includes a synthetic `weird_future_event` to lock that contract in.
- Pre-cancellation handled deterministically before the StreamReader is constructed — avoids the "stream not readable" trap. Test `PreCancelledToken_ThrowsOperationCanceledNotTransportException` covers it.
- HTTP open-time error mapping is solid: status code, reason phrase, and a body excerpt with explicit `…[truncated]` marker. Three tests (401, 500, 400-with-4KB-body) lock the contract.
- `MidStreamErrorEvent_ThrowsLlmTransportException` correctly asserts that fragments yielded BEFORE the error are preserved — that's the right semantics for an `IAsyncEnumerable`.
- `FrameSplitAcrossChunks_StillReassembles` is the right kind of test: it confirms the parser is robust to network-realistic byte boundaries, not just frame-aligned reads.
- `OperationCanceledException` is propagated unchanged (not wrapped in `LlmTransportException`) — matches the contract's "cancellation is not a transport failure" semantics.
- HttpClient ownership semantics correct: owned client disposed exactly once via `_disposed` guard; borrowed client never disposed.
- Constructor argument validation (null/whitespace api key, null model, null httpClient) matches `AnthropicTransport`'s pattern — consistent with the existing adapter.
- xmldoc on the class enumerates the contract verbatim (event types, error mapping, cancellation semantics) — good defensive documentation.

### Notes for follow-up (out of scope for this PR)

- Wiring into `LlmProviderFactory` (per #149 out-of-scope).
- The dispose-wake test is the only thing that nudges this from prototype toward MVP confidence — would recommend before this transport is wired into production paths.

**Verdict**: APPROVE

Issue #160, opened 2026-03-29
Vision: #56 ConversationRegistry accesses GameSession.Interest but it is private — needs public accessor


decay256:
## Concern
The `ConversationRegistry` spec (#56) directly accesses `entry.Session.Interest.Current`, `entry.Session.Interest.GetState()`, and `entry.Session.Interest.Apply()`. However, `GameSession._interest` is a `private readonly` field with no public accessor. The registry cannot read or modify interest without a public property.

This was flagged in the Sprint 7 vision review (recommendation #2: 'Add `public int CurrentInterest` property to GameSession as part of Wave 0') but was NOT incorporated into the #139 spec or any other issue.

## Why it matters
Without a public interest accessor on `GameSession`, issue #56 (ConversationRegistry) **cannot compile** — its ghost check, fizzle check, and interest decay logic all require reading and modifying the session's interest state.

## What needs to happen
1. Add a public property to `GameSession` that exposes the `InterestMeter` (either `public InterestMeter Interest { get; }` for full access, or a narrower API like `public int CurrentInterest` + `public InterestState CurrentState` + a method for decay)
2. This should be added as part of #139 Wave 0 (trivial one-liner) or as a prerequisite for #56
3. Update the #139 spec to include this property

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

PR #760, opened 2026-04-25
feat(setup-stream): OpenAiStreamingTransport for OpenAI/OpenRouter/Groq SSE (pinder-web #150)

decay256:
Implements `Pinder.LlmAdapters.OpenAi.OpenAiStreamingTransport : IStreamingLlmTransport, IDisposable` for any OpenAI-compatible chat completions endpoint (OpenAI, OpenRouter, Groq, Together, Ollama, ...).

Cross-ref: decay256/pinder-web#150 (epic decay256/pinder-web#148).

## What

- `OpenAiStreamingTransport` mirrors `OpenAiTransport`'s constructors (`apiKey`, `baseUrl`, `model`, optional injected `HttpClient`).
- POSTs `{baseUrl}/v1/chat/completions` with `"stream": true`, `Authorization: Bearer <key>`, `Accept: text/event-stream`.
- Parses SSE: `data: <json>` frames followed by blank lines, terminator `data: [DONE]`.
- Yields **only** non-empty `choices[0].delta.content` strings.
- Ignores: role-only first delta, `tool_calls` deltas, finish_reason chunks (empty content), SSE comment lines (`:` prefix), unknown SSE fields (`event:`, `id:`, `retry:`).
- Honours `CancellationToken` (pre-send and between SSE lines).
- Error mapping (parity with #149 Anthropic):
  - HTTP non-2xx → `LlmTransportException` (status + truncated body in message)
  - Mid-stream `{"error":{...}}` frame → `LlmTransportException`
  - Malformed SSE JSON → `LlmTransportException`
  - `OperationCanceledException` propagates unchanged.

## Why netstandard2.0 needs a new package

`Pinder.LlmAdapters` targets `netstandard2.0`/C# 8.0. `IAsyncEnumerable<T>` and `[EnumeratorCancellation]` are not in that BCL surface, so this PR adds `Microsoft.Bcl.AsyncInterfaces 8.0.0` (matching what `Pinder.Core` already does for `IStreamingLlmTransport`).

## Tests (16 new, all green)

Located in `tests/Pinder.LlmAdapters.Tests/OpenAi/OpenAiStreamingTransportTests.cs` using a fake `HttpMessageHandler`:

- `YieldsContentFragmentsInOrder`
- `RoleOnlyFirstFrame_IsIgnored`
- `DoneSentinel_TerminatesStreamCleanly` (post-`[DONE]` frames ignored)
- `FinishReasonChunkWithEmptyDelta_DoesNotYield`
[...]

Issue #148, opened 2026-03-29
Vision: ~12 open vision concerns are resolved by #139 scope — should be closed to reduce noise


decay256:
## Concern
#139 (Wave 0 Prerequisites) addresses infrastructure gaps raised across Sprints 1-5. The following open vision concerns are now resolved by #139's scope but remain OPEN, creating confusion for implementers who see 30+ open VCs:

**Fully resolved by #139:**
- #58 — StatBlock immutability → SessionShadowTracker
- #64 — TrapState.HasActive → added in #139
- #65 — RollEngine fixed-DC overload → ResolveFixedDC in #139
- #67 — IGameClock injectable → defined in #139
- #79 — InterestMeter configurable starting value → constructor overload in #139
- #82 — GameSession injection points → GameSessionConfig in #139
- #128 — Shadow mutation architecture → SessionShadowTracker in #139
- #129 — RollResult.Total excludes externalBonus → Resolve parameter in #139
- #130 — Wave 0 prerequisites needed → IS #139
- #136 — IsSuccess ignores ExternalBonus → FinalTotal in #139
- #137 — No Wave 0 implementation issues → #139 is the Wave 0

**Partially resolved (need verification after #139 ships):**
- #71 — Chaos >= high threshold → clarified in #55 body as base stat >= 4
- #74 — Horniness roll vs shadow stat → clarified in #51 body
- #81 — PlayerResponseDelay time source → clarified in #55 body

## Why it matters
30+ open VCs makes it hard to identify genuinely unresolved concerns. Implementers waste time reading stale VCs.

## What needs to happen
- [ ] After #139 is merged, close #58, #64, #65, #67, #79, #82, #128, #129, #130, #136, #137 with a comment referencing #139
- [ ] Close #71, #74, #81 with comments referencing the updated issue bodies
- [ ] This is a housekeeping task, not a code task

**Role**: technical-writer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

PR #762, opened 2026-04-25
test(llm-adapters): cover mid-blocking-read cancel + mid-stream IOException for streaming transports (pinder-web #160)

decay256:
Closes pinder-web#160 (CODE change in pinder-core; issue tracked in pinder-web).

## Summary

Closes the two coverage gaps flagged in PR #761 (Anthropic) and PR #760 (OpenAI) reviewer feedback. Adds 4 new tests (2 per transport) and fixes a real bug surfaced by one of them.

## Tests added (4)

Both transports now have these two tests:

1. **`SendStreamAsync_CancellationDuringBlockingRead_DisposesStreamAndThrowsOperationCanceled`**
   A fake `BlockingStream` emits one valid frame, then blocks in `ReadAsync` until disposed. The test starts enumerating, observes the first chunk, then cancels the token. Asserts:
   - Enumeration throws `OperationCanceledException` (NOT `LlmTransportException`).
   - The underlying stream is `Disposed == true` (the dispose-on-cancel registration actually fired).
   - This is the scenario the cancellation registration exists for; without it, the test hangs forever.

2. **`SendStreamAsync_MidStreamIOException_WrapsInLlmTransportException`**
   A fake `IoFaultStream` emits one valid frame, then throws `IOException` on the next read. Asserts:
   - Enumeration throws `LlmTransportException`.
   - `LlmTransportException.InnerException` is the original `IOException` (message preserved).

This is exactly AC-1 and AC-2 from issue #160 — both applied to `AnthropicStreamingTransportTests` and `OpenAiStreamingTransportTests`.

## Bug fix surfaced by the new tests

`OpenAiStreamingTransport` was missing the dispose-on-cancel `cancellationToken.Register(...)` hook that the Anthropic transport already had. Concretely:

- `ReadLineAsync` on `netstandard2.0` does not accept a `CancellationToken`.
- The Anthropic transport works around this by registering a callback that calls `stream.Dispose()` on cancellation, which wakes the blocked read with `ObjectDisposedException` (or `IOException` on some platforms).
- The OpenAI transport had a `ReadLineAsync` loop that only checked the token between lines, with no wake-up mechanism. A cancellation while parked inside `ReadLineAsync` would never interrupt the read — verified directly: the new mid-blocking-read test hangs against the original code.

The fix:
- Adds `cancellationToken.Register(s => stream.Dispose())` around the `using` for the response stream in `StreamCore`, mirroring the Anthropic pattern.
- Adds `ObjectDisposedException` and `IOException` catch arms in `ParseSseAsync` that re-throw as `OperationCanceledException` when the token has fired.
- Adds the same pre-cancellation `ThrowIfCancellationRequested()` guards before `ReadAsStreamAsync` and `StreamReader` construction to keep behaviour deterministic for pre-cancelled tokens.

This is the smallest possible patch that makes cancellation reliable end-to-end. Public API and happy-path semantics are unchanged.

## Verification

```
$ dotnet test tests/Pinder.LlmAdapters.Tests/Pinder.LlmAdapters.Tests.csproj \
    --filter "FullyQualifiedName~StreamingTransport" 2>&1 | tail -3
A total of 1 test files matched the specified pattern.
Passed!  - Failed:     0, Passed:    37, Skipped:     0, Total:    37, Duration: 1 s
```

Streaming transport suite: **17 + 16 = 33 → 17+2 + 16+2 = 37 passing** (delta exactly +4, no regressions).

Full `Pinder.LlmAdapters.Tests`:
- Before: `Failed: 63, Passed: 880, Skipped: 9, Total: 952`
- After:  `Failed: 63, Passed: 884, Skipped: 9, Total: 956`
- Delta: +4 passed, +0 failed (the 63 pre-existing failures are unrelated to this change).

## Files changed

- `src/Pinder.LlmAdapters/OpenAi/OpenAiStreamingTransport.cs` — dispose-on-cancel fix (~36 lines, no public API change).
- `tests/Pinder.LlmAdapters.Tests/Anthropic/AnthropicStreamingTransportTests.cs` — 2 new tests + `BlockingStream`/`IoFaultStream`/`StreamHandler` helpers.
- `tests/Pinder.LlmAdapters.Tests/OpenAi/OpenAiStreamingTransportTests.cs` — 2 new tests + same helpers.

## Out of scope

- Mid-stream resumption (called out as future work in #160 and #148).
- Cleaning up the 63 pre-existing test failures in this project.

## Refs

- pinder-web#160 — coverage gaps issue
- pinder-web#148 — Sprint B epic
- pinder-core PR #761 (Anthropic streaming transport, #149)
- pinder-core PR #760 (OpenAI streaming transport, #150)


decay256 (commented):
## QA Verdict: **PASS**

**Issue**: pinder-web #160 — coverage gaps for streaming transports (mid-blocking-read cancel + mid-stream IOException wrap).

### Targeted run
```
dotnet test tests/Pinder.LlmAdapters.Tests/ --filter "FullyQualifiedName~StreamingTransport"
Passed!  - Failed: 0, Passed: 37, Skipped: 0, Total: 37, Duration: 1 s
```

### Full LlmAdapters suite
```
Failed: 63, Passed: 884, Skipped: 9, Total: 956
```
The 63 failures are **pre-existing on `origin/main`** (verified by running `--filter ParseDialogueOptionsTests` on `8c896cd` → 8/13 fail). All failures are in `DialogueOptions` parsing / prompt-template tests, completely orthogonal to the streaming transports. **Not introduced by this PR.**

### AC mapping → tests
| AC | Test(s) | Result |
|----|---------|--------|
| Cancel mid-blocking-read → OCE, stream disposed, no LlmTransportException | `SendStreamAsync_CancellationDuringBlockingRead_DisposesStreamAndThrowsOperationCanceled` (Anthropic + OpenAi) | ✅ PASS |
| Mid-stream `IOException` → `LlmTransportException` wrapping inner `IOException` | `SendStreamAsync_MidStreamIOException_WrapsInLlmTransportException` (Anthropic + OpenAi) | ✅ PASS |
| Both tests applied to BOTH transports | 4 new test methods, 2 per transport | ✅ PASS |

The mid-blocking-read tests use a real `BlockingStream` parked inside `ReadLineAsync`, scheduling cancellation only after the first frame is yielded — i.e. the iterator is genuinely parked in the blocking read when the token fires. They assert `stream.Disposed == true` AND `IsNotType<LlmTransportException>(caught)`. This is exactly the hard path the issue calls out as untested.

### Production-code change review
`src/Pinder.LlmAdapters/OpenAi/OpenAiStreamingTransport.cs` (+34/-1) is a **legitimate bug fix surfaced by the new tests**, mirroring the dispose-on-cancel pattern that already exists in `AnthropicStreamingTransport`:
- Pre-cancel guard (`ThrowIfCancellationRequested`) before `ReadAsStreamAsync` to surface OCE rather than ArgumentException.
- `OperationCanceledException` re-thrown unwrapped from the response-body open path.
- `cancellationToken.Register(s => ((Stream)s!).Dispose())` registration on the response stream, scoped via `using`.
- New `catch (ObjectDisposedException) when (cancellationToken.IsCancellationRequested)` and `catch (IOException ioex) when (cancellationToken.IsCancellationRequested)` arms in the read loop, both calling `ThrowIfCancellationRequested()` so cancellation surfaces as OCE.

This is a sound, narrowly-scoped fix. The catch arms are guarded on `cancellationToken.IsCancellationRequested` so non-cancel `IOException`s still flow into the existing `LlmTransportException` wrap path. Symmetric with Anthropic transport. Sane.

### Coverage gap closed
Issue #160 explicitly noted the catch arms for `ObjectDisposedException`/`IOException` were "speculative" because no test parked the iterator inside `ReadLineAsync`. The new `BlockingStream` + delayed `cts.Cancel()` pattern exercises exactly that hard path. Gap closed.

### Constraints respected
No push, no merge.

### Verdict
**PASS** — all 4 new tests pass, AC fully covered, production change is a real bug fix with a clean, symmetric implementation, no test regressions introduced.

decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

### Blocking
_None._

### Non-blocking

- **`OpenAiStreamingTransport.cs:225-232` (and Anthropic counterpart)** — In the `IOException ioex when (cancellationToken.IsCancellationRequested)` arm, the `throw new LlmTransportException(...)` after `ThrowIfCancellationRequested()` is dead code: the `when` clause guarantees the token is cancelled, so the preceding `ThrowIfCancellationRequested()` always throws OCE. Cosmetic; matches the Anthropic pattern, so leaving it is defensible for symmetry. Either drop it as `throw; // unreachable` (mirroring the ODE arm immediately above) or add a comment marking it unreachable.
- **`OpenAiStreamingTransport.cs:170-178`** — The `using (responseStream)` and `using (cancellationToken.Register(...))` are stacked correctly (registration is disposed *before* the stream, so the dispose-on-cancel callback can't fire on an already-disposed stream during teardown), but the ordering is load-bearing and worth a one-line comment for the next reader. Anthropic uses an explicit `var ctRegistration = ...` which makes the intent clearer — consider matching that style for consistency with #761.

### Positive notes

- The dispose-on-cancel hook in `OpenAiStreamingTransport` is a faithful port of the Anthropic pattern: same pre-cancel guard before stream construction, same registration ordering (registration disposed before stream), same `ObjectDisposedException`/`IOException` catch arms gated by `cancellationToken.IsCancellationRequested`. Both arms call `ThrowIfCancellationRequested()` first so a non-cancelled ODE/IOException still falls through to the generic transport-wrap.
- `BlockingStream` is well-built: `TaskCompletionSource<int>` with `RunContinuationsAsynchronously` (correctly avoids a deadlock if the callback resumes on a sync context); the dispose path uses a `-1` sentinel that surfaces as `ObjectDisposedException`, matching what a real disposed stream would throw; `Disposed` flag exposed for assertion.
- `IoFaultStream` is minimal and surgical — emits a known frame, throws on next read, no timing dependency.
- Test assertions correctly model the contract: `IsAssignableFrom<OperationCanceledException>` (allows `TaskCanceledException`) **and** `IsNotType<LlmTransportException>` directly contradicts a wrap-as-transport regression. AC-2 test asserts `InnerException` type **and** message preserved — robust.
- Flake check: 5×2 = 10 runs of the two new tests, all pass, durations 17–108 ms. The 50 ms cancellation delay is generous enough that the iterator is parked inside `ReadLineAsync` when the token fires, but small enough that the test isn't slow. No timing-based flakes observed.
- Full streaming-transport suite: 37/37 pass (was 33; +4 as advertised).
- AC-1 and AC-2 from issue #160 are both met for **both** transports.
- Bug fix is in scope: the issue explicitly says "if a test surfaces a real bug in the dispose-on-cancel path, fix it and document" — done, with a clear PR-body explanation.

### Verification

```
$ dotnet test tests/Pinder.LlmAdapters.Tests/ --filter "FullyQualifiedName~StreamingTransport" | tail -1
Passed!  - Failed:     0, Passed:    37, Skipped:     0, Total:    37, Duration: 1 s

$ for i in $(seq 1 5); do dotnet test ... --filter "...CancellationDuringBlockingRead|...MidStreamIOException" | tail -1; done
Passed!  - Failed:     0, Passed:     4, Skipped:     0, Total:     4, Duration: 36 ms
Passed!  - Failed:     0, Passed:     4, Skipped:     0, Total:     4, Duration: 17 ms
Passed!  - Failed:     0, Passed:     4, Skipped:     0, Total:     4, Duration: 22 ms
Passed!  - Failed:     0, Passed:     4, Skipped:     0, Total:     4, Duration: 50 ms
Passed!  - Failed:     0, Passed:     4, Skipped:     0, Total:     4, Duration: 108 ms
```

**Verdict**: APPROVE

PR #117, opened 2026-03-28
[Issue #78] TurnResult expansion — add RiskTier enum and 7 new fields

decay256:
Fixes #78

## What was done
Added `RiskTier` enum (Safe/Medium/Hard/Bold) to `Pinder.Core.Rolls` and expanded `TurnResult` with 7 new optional fields: `ShadowGrowthEvents`, `ComboTriggered`, `CallbackBonusApplied`, `TellReadBonus`, `TellReadMessage`, `RiskTier`, `XpEarned`. All new constructor parameters are optional with safe defaults, preserving backward compatibility.

## Files changed
- `src/Pinder.Core/Rolls/RiskTier.cs` — **New**: RiskTier enum
- `src/Pinder.Core/Conversation/TurnResult.cs` — Expanded with 7 new properties + optional constructor params
- `tests/Pinder.Core.Tests/TurnResultExpansionTests.cs` — **New**: 15 tests covering defaults, population, null-coalescing, edge cases

## How to test
```bash
dotnet test
```
All 130 tests pass (115 existing + 15 new). Zero warnings.

## Deviations from contract
None — implementation matches the spec in `docs/specs/issue-78-spec.md` exactly.

## DoD Evidence
**Branch:** issue-78-turnresult-expansion-add-fields-for-new-
**Commit:** b86e970
**Tests:** Passed! - Failed: 0, Passed: 130, Skipped: 0, Total: 130

PR #249, opened 2026-04-01
[Issue #241] Write spec document: Bug: Legendary fail delivery generates wrong character voice (opponent instead of player)

decay256:
Fixes #241

## DoD Evidence
**Branch:** issue-241-write-spec-document-bug-legendary-fail-d
**Commit:** 48e878d

## Summary
Spec document for issue #241 covering the legendary fail delivery voice bug. The spec details:
- Root cause: `DeliverMessageAsync` sends both character system prompts; failure template lacks player identity framing
- Fix: 5 files changed across `Pinder.Core` and `Pinder.LlmAdapters`
- New method: `CacheBlockBuilder.BuildPlayerOnlySystemBlocks`
- Template changes: player identity framing in `FailureDeliveryInstruction`
- Wiring: `GameSession` populates `PlayerName`/`OpponentName`/`CurrentTurn` on delivery and opponent contexts
- Full edge cases, error conditions, and acceptance criteria documented

**Spec file:** `docs/specs/issue-241-spec.md`

PR #765, opened 2026-04-26
fix(#117): expose ComputeRiskTier as public static

decay256:
## Why

`DialogueOptionDto.From` in pinder-web (`src/Pinder.GameApi/Models/TurnDtos.cs`) re-implements the risk-tier threshold ladder (`need ≤7 Safe`, `≤11 Medium`, `≤15 Hard`, `≤19 Bold`, else `Reckless`) instead of calling the engine. Today the numbers agree, but if a future tweak changes the engine thresholds, the option-preview UI would silently lie.

This PR makes the canonical helper accessible so the parent repo can drop its duplicate.

## What

- `Pinder.Core.Rolls.RollResult.ComputeRiskTier(int dc, int statModifier, int levelBonus)`: `private static` → `public static`.
- XML doc note added: "canonical formula; DTOs/UI MUST call this rather than re-implement (#117)".
- No engine math change; signature & behaviour identical.

## Risk

Trivial. No call-site change inside this submodule. Build clean (0 errors).

## Pairs with

- decay256/pinder-web fix(#117): use engine's ComputeRiskTier in DialogueOptionDto (PR opened in parent)

### DIFF
--- a/src/Pinder.Core/Rolls/RollResult.cs
+++ b/src/Pinder.Core/Rolls/RollResult.cs
@@ -110,8 +110,9 @@ namespace Pinder.Core.Rolls
 
         /// <summary>
         /// Compute risk tier from the "need" value: dc - (statMod + levelBonus).
+        /// Canonical formula; DTOs/UI MUST call this rather than re-implement (#117).
         /// </summary>
-        private static RiskTier ComputeRiskTier(int dc, int statModifier, int levelBonus)
+        public static RiskTier ComputeRiskTier(int dc, int statModifier, int levelBonus)
         {
             int need = dc - (statModifier + levelBonus);
 


decay256 (commented):
## Code Review

**Maturity level:** mvp
**Blocking issues:** 0
**Non-blocking issues:** 0

### Positive notes

- One-line visibility change. Method body and signature are unchanged.
- XML doc note explicitly tells future readers (and DTO authors) that this is the canonical formula and they MUST call it rather than re-implement.
- Used by `decay256/pinder-web` PR #249 — landing this first is the correct merge order.

### Verification

- `dotnet build pinder-core/src/Pinder.Core` → 0 errors.
- Existing engine tests all pass against the parent's bumped pointer.

**Verdict:** APPROVE

Issue #255, opened 2026-04-01
Vision: #242 has no spec document — architect phase was skipped or mislabeled


decay256:
## Concern
Issue #242 (shadow taint injection) has no specification document at `docs/specs/issue-242-spec.md`. PR #249 was titled as the #242 spec but actually contained the #240 spec (`docs/specs/issue-240-spec.md`) and the sprint 10 vision review.

#242 requires non-trivial changes across two projects:
- `Pinder.LlmAdapters`: new `SessionDocumentBuilder` parameters, new `PromptTemplates` taint constants, adapter wiring
- `Pinder.Core`: GameSession shadow threshold passing to DeliveryContext/OpponentContext

Without a spec, the implementer must infer the exact taint text for 6 shadow stats × 3 tiers, the `SessionDocumentBuilder` API changes, and the GameSession wiring — all from a 4-line AC.

## Why it matters
Underspecified issues produce incomplete implementations. The #242 AC mentions only Madness and Dread at T1. There are 6 shadow stats (Dread, Madness, Fixation, Denial, Cringe, Loneliness) each with 3 threshold tiers. The implementer needs explicit taint text for each.

## What needs to happen
- [ ] Write `docs/specs/issue-242-spec.md` before implementation begins
- [ ] Spec must include: taint text for all shadow stats at each tier, SessionDocumentBuilder method signature changes, GameSession wiring for DeliveryContext/OpponentContext

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

PR #196, opened 2026-03-29
[Issue #50] Tells — implement §15 opponent tell detection and hidden roll bonus

decay256:
Fixes #50

## What was implemented

Implemented the §15 tell detection mechanic in GameSession:

- **_activeTell field**: Stores the tell from the most recent OpponentResponse.DetectedTell
- **StartTurnAsync**: Sets HasTellBonus=true on DialogueOptions whose stat matches the active tell
- **ResolveTurnAsync**: Computes tellBonus (+2 when stat matches), passes it through externalBonus to RollEngine.Resolve
- **TurnResult**: Populates TellReadBonus (0 or 2) and TellReadMessage on match
- **Consumption**: Tell is cleared after one turn in ResolveTurnAsync; also cleared by Read/Recover/Wait actions

## How to test

```bash
dotnet test --filter TellBonusTests
```

17 tests cover:
- AC1: Tell stored from opponent response
- AC2: +2 applied via externalBonus on matching stat
- AC4: TellReadBonus/TellReadMessage populated correctly (match and no-match)
- AC5: HasTellBonus flag set on matching options
- Edge cases: no tell, different stat, consumed after one turn, overwrite, first turn, nat20, nat1, still-fails, miss-to-hit, Read/Wait clearing

## Test results
- 847 passed, 1 pre-existing failure (ComboGameSessionSpecTests.AC4_Integration_TripleBonusAppliedAsExternalBonus — FixedDice exhaustion, unrelated)
- Build clean: 0 warnings, 0 errors

## Deviations from contract
None

## Files changed
- `src/Pinder.Core/Conversation/GameSession.cs` — tell logic in StartTurnAsync and ResolveTurnAsync
- `tests/Pinder.Core.Tests/TellBonusTests.cs` — 17 new tests

Issue #275, opened 2026-04-02
Vision: #267 Catastrophe should also activate trap on Legendary (Nat 1) tier


decay256:
## Concern
Issue #267 adds trap activation for Catastrophe (miss ≥10), but rules §5 says Legendary (Nat 1) is the worst tier — it includes everything Catastrophe does plus more. Currently, Legendary (Nat 1) fires before the miss-margin check in `ResolveFromComponents()` and skips trap activation entirely.

Rules §5 tier hierarchy: Legendary > Catastrophe > TropeTrap > Misfire > Fumble. If Catastrophe activates traps, Legendary should too.

Current code flow:
```
if (usedRoll == 1) → tier = Legendary  // exits before miss check, NO trap activation
else if (miss <= 9) → tier = TropeTrap // activates trap ✅
else → tier = Catastrophe              // #267 will add trap activation ✅
```

After #267, Legendary is the ONLY failure tier that doesn't activate a trap, despite being the worst outcome.

## Why it matters
Legendary failures (Nat 1) are supposed to be devastating — trap + shadow growth + extra penalty. Without trap activation, Nat 1 is mechanically less punishing than missing by 10+ (Catastrophe), which is counter-intuitive and rules-incorrect.

## What needs to happen
When implementing #267, also add trap activation for the Legendary (Nat 1) branch:
```csharp
if (usedRoll == 1)
{
    tier = FailureTier.Legendary;
    // Activate trap (same as TropeTrap/Catastrophe)
    if (!attackerTraps.IsActive(stat))
    {
        newTrap = trapRegistry.GetTrap(stat);
        if (newTrap != null) attackerTraps.Activate(newTrap);
    }
}
```

## Acceptance Criteria
- [ ] Legendary (Nat 1) activates a trap if one is defined for the stat
- [ ] Test: Nat 1 roll activates trap
- [ ] All three trap-activating tiers (TropeTrap, Catastrophe, Legendary) tested

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision concern from early sprint coordination. Architecture has evolved significantly since this was filed.

PR #766, opened 2026-04-26
feat(#255): trap display_name + summary fields

decay256:
## Summary

Add player-facing `display_name` (Title-Case) and `summary` (one-line flavour) to trap definitions so the web UI can show the specific trap name and description in the failure-tier label and activated-trap line — instead of a generic "Trope Trap" with a kebab-case id like `cringe`.

Tracks issue: decay256/pinder-web#255

This PR is the **submodule half** of that bundle. The parent repo PR (decay256/pinder-web) bumps the pinder-core pointer and wires the new fields through the C# DTO + frontend types + render.

### Data
- `data/traps/traps.json`: `display_name` + `summary` on all 6 canonical traps (cringe / creep / overshare / unhinged / pretentious / spiral).
- `data/traps/trap-schema.json`: optional fields with safe defaults (so older data files without the fields still validate).

### Engine
- `src/Pinder.Core/Traps/TrapDefinition.cs`: new `DisplayName` and `Summary` properties.
  - Constructor takes both as **optional trailing parameters** so existing call sites keep compiling unchanged.
  - `DisplayName` falls back to `Id` when absent or empty.
  - `Summary` defaults to empty string.
- `src/Pinder.Core/Data/JsonTrapRepository.cs`: parses both fields via `obj.GetString(key, defaultValue)`.

### Tests
- `tests/Pinder.Core.Tests/JsonTrapRepositoryDisplayNameSummaryTests.cs`:
  - canonical data has non-empty `display_name` + `summary` on all 6 traps
  - per-stat spec values for `display_name` (Cringe / Creep / Overshare / Unhinged / Pretentious / Spiral)
  - one full-text `summary` spot-check (cringe)
  - legacy JSON without the fields still loads (`DisplayName → Id`, `Summary → ""`)
  - explicit JSON values parse correctly

## Wire-shape compatibility

Wire-shape change is **additive only**: no existing field renamed or removed. Backwards-compatible with all #117 / #196 consumers; the existing `id` field stays on the wire as the primary key, and the deprecated `activated_trap_name` alias keeps working.

## Test evidence

```
$ dotnet test tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj --no-build --filter "FullyQualifiedName~Trap|FullyQualifiedName~JsonTrapRepository"
Passed!  - Failed:     0, Passed:   176, Skipped:     0, Total:   176
```

Pre-existing test failures in `Pinder.Rules.Tests`, `Pinder.LlmAdapters.Tests`, and a handful of `CharacterLoaderSpecTests` are unrelated to this change (they fail identically on the unmodified base commit when run from this clone — environmental data-path issue, not something this PR touches).

## Acceptance against #255

- [x] All 6 traps in `traps.json` have `display_name` + `summary`.
- [x] Engine type carries the new fields.
- [x] Backwards-compatible default behaviour for legacy data.
- [ ] **DTO + frontend wiring + render** — handled in the parent repo PR.

### DIFF
--- a/data/traps/trap-schema.json
+++ b/data/traps/trap-schema.json
@@ -11,6 +11,16 @@
         "type": "string",
         "description": "Unique identifier for the trap (e.g. 'cringe', 'creep')."
       },
+      "display_name": {
+        "type": "string",
+        "description": "Player-facing Title-Case name of the trap (e.g. 'Cringe', 'Creep'). Used in the failure-tier label and the activated-trap line. Optional; falls back to id when absent. (#255)",
+        "default": ""
+      },
+      "summary": {
+        "type": "string",
+        "description": "Short one-sentence player-facing description of the trap's flavour. Distinct from llm_instruction (which is internal). Rendered under the activated-trap line. Optional; empty string omits the line. (#255)",
+        "default": ""
+      },
       "stat": {
         "type": "string",
         "enum": ["charm", "rizz", "honesty", "chaos", "wit", "self_awareness"],
--- a/data/traps/traps.json
+++ b/data/traps/traps.json
@@ -1,6 +1,8 @@
 [
   {
     "id": "cringe",
+    "display_name": "Cringe",
+    "summary": "You're aware of how you're coming across, which is making it worse.",
     "stat": "charm",
     "effect": "disadvantage",
     "effect_value": 0,
@@ -11,6 +13,8 @@
   },
   {
     "id": "creep",
+    "display_name": "Creep",
+    "summary": "An accidental 'agenda' quality is leaking into your messages.",
     "stat": "rizz",
     "effect": "stat_penalty",
     "effect_value": 2,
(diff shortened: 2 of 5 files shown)


decay256 (commented):
## Code Review

**Maturity level:** mvp
**Blocking issues:** 0
**Non-blocking issues:** 0

### Positive notes

- Added `display_name` + `summary` to **all 6** traps in `traps.json` (the issue body said 4 but there are 6 in production data — author caught this).
- `trap-schema.json` updated: both fields optional, allowing schema-validated old data files to keep validating.
- `TrapDefinition.cs` constructor parameters added as **optional trailing args** so existing test fixtures with 8 positional args don't break. Backward-compatible.
- `JsonTrapRepository.cs` parser uses `GetString(key, default)` overload — silent fallback when fields are missing.
- 12 new tests in `JsonTrapRepositoryDisplayNameSummaryTests`.
- Used by parent PR #275; landing this first is the correct merge order.

**Verdict:** APPROVE

Issue #310, opened 2026-04-03
Missing: Madness T3 (≥18) should replace one dialogue option with unhinged text


decay256:
## Context
Rules §7 shadow thresholds: 'Madness at 18+: One option/turn replaced with unhinged text.'

Denial T3 (remove Honesty options) and Fixation T3 (force last stat) are implemented in `GameSession.StartTurnAsync()`. Madness T3 is not.

## Where to implement
`GameSession.StartTurnAsync()` after the existing Denial T3 and Fixation T3 blocks (roughly line ~310):

```csharp
// Madness T3: replace one random option with unhinged replacement marker
if (shadowThresholds.TryGetValue(ShadowStatType.Madness, out int madnessTier)
    && madnessTier >= 3
    && options.Length > 0)
{
    // Pick a random option to mark as unhinged
    int unhingedIdx = _dice.Roll(options.Length) - 1;
    var o = options[unhingedIdx];
    options[unhingedIdx] = new DialogueOption(
        o.Stat, o.IntendedText, o.CallbackTurnNumber,
        o.ComboName, o.HasTellBonus, o.HasWeaknessWindow,
        isUnhingedReplacement: true);  // requires adding this field to DialogueOption
}
```

## Changes needed
1. Add `IsUnhingedReplacement` bool to `DialogueOption` (default false)
2. Add the T3 block in `GameSession.StartTurnAsync()`
3. Pass `IsUnhingedReplacement` flag in `DialogueContext` or as part of the option — so the LLM adapter can request appropriately unhinged text for that option

## Acceptance Criteria
- [ ] `DialogueOption` has `IsUnhingedReplacement` bool
- [ ] At Madness T3 (≥18): exactly one option is marked `IsUnhingedReplacement = true`
- [ ] At Madness T2 or lower: no options marked unhinged
- [ ] The LLM prompt notes which option slot should be unhinged
- [ ] Tests verify T3 marks one option, T2 marks none
- [ ] Build clean

PR #305, opened 2026-04-03
[Issue #273] Write tests: Missing: Madness T3 (≥18) replaces one option with unhinged text (§7)

decay256:
Fixes #273

## DoD Evidence
**Branch:** issue-273-write-tests-missing-madness-t3-18-replac
**Commit:** 4dd247c

## Test Summary
18 tests covering Madness T3 unhinged option behavior:

### AC Coverage
- **AC1**: Madness≥18 marks exactly one option IsUnhinged (4 tests)
- **AC3**: Madness<18 marks no options (4 tests including boundary at 17)
- **AC3 deterministic**: Dice controls which index is marked (3-value Theory)
- **AC4**: DialogueOption backward compat (IsUnhinged defaults false)

### Edge Cases
- Single option available → becomes unhinged
- No shadow tracker → no crash, no marking
- Madness T3 + Horniness T3 → flag preserved after Rizz conversion
- Madness T3 + Fixation T3 → both effects apply
- Stat preservation on unhinged option

All 18 tests pass.

PR #767, opened 2026-04-28
feat(#305): TurnSnapshot.ConversationEntry carries per-layer text_diffs[]

decay256:
Per AGENTS.md schema-discipline rule (every player-visible `GameSession` field MUST be reflected in `TurnSnapshot`), `TextDiffs[]` — produced by `GameSession.ResolveTurnAsync` on `TurnResult.TextDiffs` — now flows into `ConversationEntry` on each turn snapshot.

Companion to **decay256/pinder-web#310** (which addresses the same issue from the GameApi side: audit row + state envelope persistence). This submodule PR is the schema-discipline counterpart.

## Changes

- `ConversationEntry` gains `TextDiffs` (`List<TextDiffSnapshot>`), defaulting to empty so legacy snapshots round-trip cleanly.
- `TextDiffSnapshot` / `TextDiffSpanSnapshot` mirror the wire `TextDiffDto` shape.
- `BuildTurnSnapshot` accepts an optional `perTurnTextDiffs` list and attaches each turn's diffs to the player entry (history index `2*(N-1)`). Opponent entries leave `TextDiffs` empty.
- `Program.cs` accumulates per-turn `TextDiffs` alongside `conversationHistory` and threads them through.
- Top-of-class "fields covered" comment block updated to list `ConversationLog + per-layer text_diffs[]`.

## Verification

End-to-end JSON round-trip verified: serialize a populated `TurnSnapshot` → `textDiffs` array present on the player entry with all three layers + spans → deserialize → all values intact.

Refs decay256/pinder-web#305.

### DIFF
--- a/session-runner/Program.cs
+++ b/session-runner/Program.cs
@@ -753,6 +753,17 @@ class Program
             ? resimTurnSnap.ConversationHistory.Select(e => (e.Sender, e.Text)).ToList()
             : new List<(string Sender, string Text)>();
 
+        // #305: parallel list of per-turn text_diffs[] for the player's
+        // delivered message. Index = turn_number - 1. When resuming from a
+        // resim snapshot we lift the diffs from the player entries (the
+        // opponent entries always carry an empty list).
+        var perTurnTextDiffs = isResimulation && resimTurnSnap != null
+            ? resimTurnSnap.ConversationHistory
+                .Where((_, idx) => idx % 2 == 0)
+                .Select(e => e.TextDiffs ?? new List<TextDiffSnapshot>())
+                .ToList()
+            : new List<List<TextDiffSnapshot>>();
+
         string lastOpponentMsg = isResimulation && resimTurnSnap != null
             ? (resimTurnSnap.ConversationHistory.LastOrDefault(e => e.Sender == player2)?.Text ?? "")
             : "";
@@ -1081,6 +1092,24 @@ class Program
             // Track conversation history for LLM agent context (#492)
             if (!string.IsNullOrEmpty(result.DeliveredMessage))
                 conversationHistory.Add((player1, result.DeliveredMessage));
+                // #305: stash this turn's TextDiffs (snapshot shape) so we can
+                // attach them to the player's history entry when building
+                // the turn snapshot below. Empty list when no layer
+                // transformed the text.
+                perTurnTextDiffs.Add(
+                    (result.TextDiffs ?? Array.Empty<Pinder.Core.Text.TextDiff>())
+                        .Select(d => new TextDiffSnapshot
+                        {
+                            Layer = d.LayerName,
+                            Before = d.Before,
+                            After = d.After,
+                            Spans = d.Spans.Select(s => new TextDiffSpanSnapshot
+                            {
+                                Type = s.Type.ToString(),
+                                Text = s.Text,
+                            }).ToList(),
+                        })
+                        .ToList());
             lastOpponentMsg = result.OpponentMessage ?? "";
             if (!string.IsNullOrEmpty(lastOpponentMsg))
                 conversationHistory.Add((player2, lastOpponentMsg));
@@ -1267,7 +1296,8 @@ class Program
                     rizzCumulativeFailureCount,
                     conversationHistory,
                     comboHistoryForSnapshot,
-                    tellSnap);
+                    tellSnap,
+                    perTurnTextDiffs);
 
                 string turnSnapPath = Path.Combine(playtestDir, $"{sessionSlug}.turn-{turn:D2}.snap.json");
                 File.WriteAllText(turnSnapPath, JsonSerializer.Serialize(turnSnap, new JsonSerializerOptions { WriteIndented = true }));
@@ -1563,7 +1593,8 @@ class Program
         int rizzCumulativeFailureCount,
         List<(string Sender, string Text)> conversationHistory,
         List<(StatType Stat, bool Succeeded)> comboHistory,
-        TellSnapshot? activeTell)
+        TellSnapshot? activeTell,
+        List<List<TextDiffSnapshot>>? perTurnTextDiffs = null)
     {
         var state = result.StateAfter;
 
@@ -1583,10 +1614,26 @@ class Program
             .Select(e => new TurnHistoryEntry { Stat = e.Stat.ToString(), Succeeded = e.Succeeded })
             .ToList();
 
-        // Conversation history
-        var convEntries = conversationHistory
-            .Select(e => new ConversationEntry { Sender = e.Sender, Text = e.Text })
-            .ToList();
+        // Conversation history (#305: attach per-turn text_diffs[] to the
+        // player's entry on each turn so the snapshot can be deserialised
+        // straight into a renderer / replay tool. Each resolved turn
+        // appends two entries — player then opponent — so player entries
+        // live at even indices and turn-N maps to history index 2*(N-1).)
+        var convEntries = new List<ConversationEntry>(conversationHistory.Count);
+        for (int i = 0; i < conversationHistory.Count; i++)
+        {
+            var (sender, text) = conversationHistory[i];
+            var entry = new ConversationEntry { Sender = sender, Text = text };
+            bool isPlayerEntry = (i % 2) == 0;
+            int turnIdx = i / 2;
+            if (isPlayerEntry
+                && perTurnTextDiffs != null
+                && turnIdx < perTurnTextDiffs.Count)
+            {
+                entry.TextDiffs = perTurnTextDiffs[turnIdx] ?? new List<TextDiffSnapshot>();
+            }
+            convEntries.Add(entry);
+        }
 
         return new TurnSnapshot
         {
--- a/session-runner/Program.cs
+++ b/session-runner/Program.cs
@@ -1091,11 +1091,16 @@ class Program
 
             // Track conversation history for LLM agent context (#492)
             if (!string.IsNullOrEmpty(result.DeliveredMessage))
+            {
                 conversationHistory.Add((player1, result.DeliveredMessage));
                 // #305: stash this turn's TextDiffs (snapshot shape) so we can
                 // attach them to the player's history entry when building
                 // the turn snapshot below. Empty list when no layer
-                // transformed the text.
+                // transformed the text. MUST stay in lock-step with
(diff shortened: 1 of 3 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 1
**Non-blocking issues**: 2

### Blocking
- **[session-runner/Program.cs:1093–1110]** Indentation/scope bug. The new `perTurnTextDiffs.Add(...)` call sits visually *inside* the single-line `if (!string.IsNullOrEmpty(result.DeliveredMessage))` block but, because the `if` has no braces, only the first statement (`conversationHistory.Add(...)`) is conditional. The `Add` to `perTurnTextDiffs` always runs. Result: when `DeliveredMessage` is empty, `conversationHistory` skips the player entry but `perTurnTextDiffs` still grows — and downstream `BuildTurnSnapshot` walks history with `turnIdx = i / 2`, so every subsequent turn's diffs land on the wrong entry. In normal play `DeliveredMessage` is never empty so this is latent, but the misleading indentation is a footgun and the invariant is undefended. Either:
  1. Add braces and put both `Add` calls inside the `if`, OR
  2. Pull `perTurnTextDiffs.Add(...)` out to column-13 (the same column as the `if` and `lastOpponentMsg = …`) so it's visibly *unconditional*, and add a comment that the empty-DeliveredMessage edge case is intentionally tolerated.

  Either way the current "looks-conditional, isn't-conditional" state should not ship.

### Non-blocking
- **[session-runner/Program.cs:760–765]** The resim path lifts diffs with `.Where((_, idx) => idx % 2 == 0)` — silently assuming every legacy snapshot has a strict player/opponent/player/opponent ordering. If a snapshot ever omitted a player turn (same edge case as the blocking issue), resim would attach diffs to the wrong turn. A defensive cross-check (e.g. assert `e.Sender == player1` for the kept entries) would make this self-diagnosing.
- **[session-runner/Snapshot/SessionSnapshot.cs:139]** `ConversationEntry.TextDiffs` defaults to a non-empty list initialiser (`= new List<TextDiffSnapshot>()`). That's friendly for legacy snapshots (deserialising a pre-#305 file leaves it empty rather than null), but it means *every* entry in the JSON output now carries `"textDiffs": []` — including opponent entries. Harmless but slightly chatty; if you want the same `WhenWritingNull` behaviour the GameApi side has, switch to `List<TextDiffSnapshot>?` + `JsonIgnore` when null. Cosmetic.

### Positive notes
- **Schema-discipline rule satisfied**: `TextDiffs` is added to `ConversationEntry` in `TurnSnapshot`, AND the top-of-class "fields covered" comment block is updated to list `ConversationLog (per-entry sender + text + per-layer text_diffs[] — issue #305)`. Exactly what AGENTS.md requires.
- **Wire-shape mirroring**: `TextDiffSnapshot` / `TextDiffSpanSnapshot` carry the same `Layer`/`Before`/`After`/`Spans` + `Type`/`Text` shape as `TextDiffDto` (pinder-web). A snapshot can be serialised straight into a renderer without any field translation.
- **Index-aligned attachment** to the player entry on each turn at history index `2*(turn-1)` matches the algorithm used by `pinder-web`'s `ActiveSession`/`GameSessionRepository`. The two repos will reconstruct the same timeline from the same data.
- **Coordination with parent PR is explicit** in the body: schema-discipline counterpart to decay256/pinder-web#310, both addressing #305. Submodule pointer dance is acknowledged.
- **Round-trip verification** mentioned in the PR body (populated TurnSnapshot serialise → parse → values intact). Worth landing as an actual unit test in a follow-up if this repo ever grows a session-runner test project.

**Verdict**: REQUEST_CHANGES (one indentation/scope bug; everything else is solid)

decay256:
Pushed [`ff9c6fb`](https://github.com/decay256/pinder-core/commit/ff9c6fb) addressing the blocking review comment.

**Fix:** wrapped the two `Add` calls in an explicit `{ }` block so `perTurnTextDiffs` only grows together with the player entry in `conversationHistory`. Picked option (1) from your suggestion — the empty-`DeliveredMessage` skip is now symmetric on both lists, which is also what `BuildTurnSnapshot` requires (it pairs the lists by player-entry index, not by raw history index, so co-indexing on the player axis is the right invariant).

**Comment expanded** at the call site to spell out the lock-step contract (so the next person who sees this code doesn't need to chase `BuildTurnSnapshot` to understand why both lines are inside the same `if`).

**Regression test:** `tests/Pinder.Core.Tests/Issue767BraceScopeTests.cs`
- happy-path: 3-turn run, diffs land on the right player entry (indices 0, 2, 4)
- empty-`DeliveredMessage` on turn 2: asserts `conversationHistory.Count(p1) == perTurnTextDiffs.Count` (the contract the fix establishes)
- counter-example: replays the original brace-less loop body and asserts it produces the desync (3 diffs vs 2 player entries) — so any future regression that re-introduces the bug fails this test loudly rather than silently misattributing diffs.

Made `BuildTurnSnapshot` `internal static` so the regression can be pinned (the project already has `InternalsVisibleTo Pinder.Core.Tests`).

```
Passed!  - Failed: 0, Passed: 3, Skipped: 0, Total: 3 - Pinder.Core.Tests.dll
```

The two non-blocking observations (resim `.Where(idx % 2 == 0)` defensive cross-check; `TextDiffs` empty-list serialisation chattiness) left for a follow-up since they're scoped beyond this PR's blocker.

decay256 (commented):
## Code Review (second pass)

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 0 (the two from the first pass are explicitly deferred and stand)

Re-review of `ff9c6fb` against the prior blocker. Fresh-eyes verification, no carry-over context.

### Prior blocker — resolved

- **`session-runner/Program.cs:1093–1118`** — both `conversationHistory.Add` and `perTurnTextDiffs.Add` are now inside an explicit `{ }` block gated by `if (!string.IsNullOrEmpty(result.DeliveredMessage))`. The misleading-indentation footgun is gone, and the invariant is now defended by tests (see below). The expanded inline comment (lines 1096–1103) names the lock-step contract that `BuildTurnSnapshot` relies on — the next reader does not need to chase the snapshot builder to understand why both `Add`s belong in the same block.
- Behaviour matches what the original code visually appeared to do: under the new code, the empty-`DeliveredMessage` skip is symmetric on both lists, which is the player-axis lock-step invariant `BuildTurnSnapshot` reads back via `turnIdx = i / 2`.

### Regression tests — sound

`tests/Pinder.Core.Tests/Issue767BraceScopeTests.cs` (3 tests, all pass on this branch):

- **`BuildTurnSnapshot_AttachesPerTurnDiffsToCorrectPlayerEntry`** — pins the read-side indexing contract: player entries at even indices carry that turn's diffs; opponent entries (odd indices) stay empty. Tests behaviour, not implementation.
- **`EmptyDeliveredMessage_LoopBody_KeepsListsLockstepOnPlayerAxis`** — replays the now-fixed loop body across 3 turns with turn 2 having an empty `DeliveredMessage`, then asserts `playerEntries == perTurnTextDiffs.Count` AND that the surviving diffs correspond to turns 1 and 3 in order. This is exactly the invariant the brace fix establishes; would have failed under the buggy version (assertion `2 == 3` would have flipped).
- **`EmptyDeliveredMessage_BuggyLoopBody_DesyncsLists`** — counter-example test that replays the *original brace-less* loop body verbatim and asserts the desync (3 diffs vs 2 player entries). I checked `BuggyApplyTurn` line-by-line against the original code: the unconditional `perTurnTextDiffs.Add(turnDiffs)` is correctly placed *outside* the brace-less `if`, faithfully replaying the bug rather than the test author's mental model of it. So a future regression that reverts the braces fails this test loudly rather than silently misattributing diffs.

No flakiness, no test-that-tests-nothing.

### Visibility change — justified

- `BuildTurnSnapshot` widened from `static` (private-by-default) to `internal static`. Required to let the test pin the indexing contract.
- `session-runner/session-runner.csproj` already carries `<InternalsVisibleTo Include="Pinder.Core.Tests" />` (line 9). No public surface area widened. `static` was already there — no regression.
- The extra `perTurnTextDiffs` parameter on `BuildTurnSnapshot` is `List<List<TextDiffSnapshot>>?` with a default of `null`, so existing call sites (and the test that doesn't supply it) keep working.

### Scope

`git diff ff9c6fb~1 ff9c6fb --stat` shows only:
- `session-runner/Program.cs` (+12 lines: braces + expanded comment + `internal` keyword)
- `tests/Pinder.Core.Tests/Issue767BraceScopeTests.cs` (new, 258 lines)

No unrelated edits crept in.

### Tests run

```
dotnet test --filter "FullyQualifiedName~Issue767BraceScopeTests"
Passed!  - Failed: 0, Passed: 3, Skipped: 0, Total: 3, Duration: 21 ms

dotnet test tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj
Failed!  - Failed: 6, Passed: 2411, Skipped: 18, Total: 2435, Duration: 5 s
```

All 6 failures are pre-existing fixture failures in `CharacterLoaderSpecTests` (5× `Load_StarterCharacter_HasValidProfile` + `Load_Gerald_MatchesSpecExactValues`) — all of them fail with `Assert.Fail("SKIPPED: Prompt directory 'design/examples' not found in repo")`. Unrelated to this PR. No new failures introduced.

### Deferred (still standing from first pass)

- Resim path `.Where((_, idx) => idx % 2 == 0)` defensive cross-check.
- `ConversationEntry.TextDiffs` empty-list serialisation chattiness.

Both correctly out-of-scope for this PR.

### Also out-of-scope (acknowledged)

The author's note that `turnIdx = i / 2` in `BuildTurnSnapshot` can still desync if a player turn is skipped *entirely* (i.e. opponent entry without a preceding player entry) is correct — the player-axis lock-step holds only as long as turns alternate strictly. That's a pre-existing assumption of the surrounding code, not a regression introduced by this PR. Worth a follow-up issue but not a blocker here.

### Positive notes

- Fix is minimal, targeted, and matches reviewer option (1).
- Comment block at the call site documents the lock-step contract explicitly — future readers will not have to reverse-engineer it.
- Counter-example test pattern (replay the buggy loop body in the test) is the right shape for "make sure a future regression that re-introduces this fails loudly". Worth keeping as a template for similar indentation/scope traps.

**Verdict**: APPROVE / CLEAR_TO_MERGE

(Posted as `--comment` because gh-cli identity matches the PR author; treat the above as an explicit approval signal.)

Issue #130, opened 2026-03-28
Vision: Continuation sprint needs Wave 0 prerequisites before feature work — SessionShadowTracker, IGameClock, RollEngine externalBonus


decay256:
## Status: ✅ RESOLVED by Issue #139 (Wave 0 Prerequisites)

SessionShadowTracker, IGameClock, RollEngine extensions, and GameSessionConfig are all defined in #139 with full contracts.

---

## Concern
Previous sprints correctly identified that prerequisite infrastructure PRs (#63 ILlmAdapter expansion, #78 TurnResult expansion) must ship before feature work. This continuation sprint has no equivalent Wave 0.

Three architectural gaps remain from open vision/arch concerns:
1. **#58/#82**: No `SessionShadowTracker`, no `IGameClock`, no `GameSessionConfig` — needed by #44, #45, #51, #54, #55, #56
2. **#85**: `RollResult.Total` excludes external bonuses — needed by #46, #47, #50 implementations
3. **#64**: `TrapState.HasActive` property doesn't exist — needed by #43 Recover action
4. **#65**: `RollEngine.Resolve` has no fixed-DC overload — needed by #43 Read/Recover

These should be bundled into 1-2 small preparatory PRs before feature implementation begins.

## Why it matters
- Without these, every feature agent will either: (a) invent their own incompatible solution, (b) hit a wall and fail, or (c) add breaking changes that conflict with parallel work
- Previous sprints proved the prerequisite-PR pattern works well (#63, #78 were clean)

## What needs to happen
Create 1-2 prerequisite issues:
- **Prerequisite A**: `SessionShadowTracker` + `IGameClock` + `GameSessionConfig` optional param on GameSession
- **Prerequisite B**: `RollEngine.Resolve` fixed-DC overload + `externalBonus` param + `TrapState.HasActive`
These ship as Wave 0 before any feature work.

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision/arch concern from early sprint coordination. Issue references old sprint issues that are closed or superseded. Architecture has evolved significantly since this was filed.

PR #768, opened 2026-04-28
fix(decay256/pinder-web#130): seed OptionFilterEngine.DrawRandomStats for test determinism

decay256:
Required by decay256/pinder-web#130. Replaces the unseeded `new Random()` in `OptionFilterEngine.DrawRandomStats` with an injectable `Random` and adds `GameSessionConfig.StatDrawRng` so `pinder-web` can plumb a seeded RNG through. Default behaviour is unchanged when no RNG is injected.

Submodule pointer bump in `pinder-web` follows in decay256/pinder-web#130's PR.

Tests:
- `Pinder.Core.Tests`: 2408 passed / 6 environmental skips (no design/examples on disk) / 18 skipped — same baseline as `origin/main` on this checkout (verified against a pristine clone).
- `Pinder.Rules.Tests`, `Pinder.LlmAdapters.Tests`: pre-existing failure counts unchanged (YAML parser / parse-defaults tests, all unrelated).

### DIFF
--- a/src/Pinder.Core/Conversation/GameSession.cs
+++ b/src/Pinder.Core/Conversation/GameSession.cs
@@ -90,6 +90,10 @@ namespace Pinder.Core.Conversation
         private readonly SessionXpRecorder _xpRecorder;
         private readonly SteeringEngine _steeringEngine;
         private readonly HorninessEngine _horninessEngine;
+        // Dedicated RNG for OptionFilterEngine.DrawRandomStats. Kept separate from
+        // the steering RNG so tests can queue exact steering values without our
+        // stat-draw shuffle consuming them (see issue #130).
+        private readonly Random? _statDrawRng;
 
         /// <summary>
         /// Creates a new GameSession with required configuration.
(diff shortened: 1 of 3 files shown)


decay256:
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 0

### Positive notes
- `OptionFilterEngine.DrawRandomStats` change is minimal and correct: optional `Random? rng` at the end of the parameter list, fallback to `new Random()` when null, preserving legacy non-deterministic behaviour for existing callers. Existing positional callers (`tests/Pinder.Core.Tests/DecomposedModuleTests.cs:139, 154`) compile and run unchanged.
- `GameSessionConfig.StatDrawRng` added at the end of the constructor parameter list with a default of `null`. All `new GameSessionConfig(...)` call sites I inspected (both repos) use named arguments, so the addition is non-breaking. XML docs explain why this is a separate knob from `SteeringRng` (so steering tests with tightly-queued `FixedRandom` aren’t perturbed by an unrelated stat-draw shuffle).
- `GameSession` plumbs the RNG through cleanly: stored as a private readonly field, used at the single `DrawRandomStats` call site, and the rest of the file is untouched.
- No behavioural change for production callers (which don’t set `StatDrawRng`); deterministic behaviour only kicks in when an RNG is explicitly injected (i.e. from `pinder-web/TestSessionStore`).

No findings on this PR independent of pinder-web#311. Approval here is conditional on pinder-web#311 landing the consumer wiring; this PR on its own is a no-op API extension.

**Verdict**: APPROVE

PR #293, opened 2026-04-02
[Issue #267] Bug: Catastrophe tier should also activate trap (rules §5)

decay256:
Fixes #267

## What Changed
`RollEngine.ResolveFromComponents()` only activated traps at TropeTrap tier (miss 6-9). Rules §5 states Catastrophe (miss 10+) should also activate a trap. Added trap activation logic to the Catastrophe branch, mirroring the existing TropeTrap logic.

## How to Test
```bash
dotnet test --filter 'Catastrophe|TropeTrap_StillActivatesTrap'
```

## Tests Added
- `Catastrophe_ActivatesTrap` — miss by 14 activates trap
- `Catastrophe_DoesNotActivateTrap_WhenAlreadyActive` — no double activation
- `TropeTrap_StillActivatesTrap` — regression test for existing behavior

## DoD Evidence
**Tests**: 1204 core + 443 adapter = 1647 passed, 0 failed
**Commit**: 748e8cf
**Deviations from contract**: none

PR #292, opened 2026-04-02
[Issue #266] Bug: FailureScale interest deltas don't match rules-v3.4 §5

decay256:
Fixes #266

## Summary
Aligned `FailureScale.GetInterestDelta()` with rules-v3.4 §5. The prototype defaults from #28 were more punishing than the rules specify.

### Changes
| Tier | Before | After (rules §5) |
|---|---|---|
| Fumble | −1 | −1 (unchanged) |
| Misfire | −2 | **−1** |
| TropeTrap | −3 | **−2** (+ trap) |
| Catastrophe | −4 | **−3** (+ trap + shadow) |
| Legendary | −5 | **−4** (+ trap + shadow +1) |

### Files Changed
- `src/Pinder.Core/Rolls/FailureScale.cs` — fixed deltas
- `tests/Pinder.Core.Tests/GameSessionTests.cs` — updated FailureScaleTests InlineData, fixed cascading interest in EndCondition and GhostTrigger tests
- `tests/Pinder.Core.Tests/Integration/FullConversationIntegrationTest.cs` — updated cascading interest assertions through 8-turn integration test

### How to Test
```bash
dotnet test tests/Pinder.Core.Tests/
```
All 1201 Core tests pass. All 443 LlmAdapters tests pass.

**Deviations from contract:** none

PR #770, opened 2026-04-28
feat(GameSession): add typed MarkEnded(GameOutcome) API

decay256:
Adds `GameSession.MarkEnded(GameOutcome outcome)` so post-game rehydration paths can mark a freshly-constructed session as ended without reflection.

```csharp
public void MarkEnded(GameOutcome outcome)
{
    _ended = true;
    _outcome = outcome;
}
```

## Why

`GameSession.RestoreState` is intentionally scoped to mid-game resimulation and does **not** touch the terminal `_ended` / `_outcome` flags. When pinder-web rehydrates an **already-finished** session from the DB, it needs to set those flags so:

1. `StartTurnAsync` throws `GameEndedException` with the correct outcome (defence-in-depth).
2. The fast-path in pinder-web's `SessionsController.GetTurnAsync` returns the persisted `EndStateDto`.

Today pinder-web (PR #292) does this via reflection on `_ended` / `_outcome`. Works, but brittle to internal renames. This PR adds the typed API; pinder-web's follow-up PR drops the reflection.

## Tests

New `Issue293_MarkEndedTests` in `tests/Pinder.Core.Tests/Conversation/`:

- `MarkEnded_SetsTerminalFlags` (theory over all three `GameOutcome` values) — asserts `IsEnded` flips and `Outcome` matches.
- `MarkEnded_StartTurnAsync_ThrowsGameEndedExceptionWithOutcome` — full integration of the contract (mark → next `StartTurnAsync` throws typed exception with the right outcome).
- `MarkEnded_OverridesUnmatched_ThrowsWithCorrectOutcome` — sanity check that a non-default outcome flows through.

```
Passed!  - Failed:     0, Passed:    68, Skipped:     0, Total:    68
```

Refs decay256/pinder-web#293.

### DIFF
--- a/src/Pinder.Core/Conversation/GameSession.cs
+++ b/src/Pinder.Core/Conversation/GameSession.cs
@@ -213,6 +213,24 @@ namespace Pinder.Core.Conversation
         public GameOutcome? Outcome => _outcome;
 
         /// <summary>
+        /// Restore an already-ended session from persisted state. Sets the
+        /// terminal flags so subsequent <see cref="StartTurnAsync"/> throws
+        /// <see cref="GameEndedException"/> with the right outcome.
+        ///
+        /// Intended for post-game replay/rehydration paths (e.g. loading a
+        /// finished session back from storage). <see cref="RestoreState"/>
+        /// targets mid-game resimulation and deliberately does not touch the
+        /// terminal flags; callers reviving an ended session must call this
+        /// in addition.
+        /// </summary>
+        /// <param name="outcome">The terminal <see cref="GameOutcome"/> the session ended with.</param>
+        public void MarkEnded(GameOutcome outcome)
+        {
+            _ended = true;
+            _outcome = outcome;
+        }
+
+        /// <summary>
         /// Conversation history as (sender, text) tuples, in emission order.
         /// Read-only snapshot view; safe to enumerate concurrently with session mutation
         /// since the underlying list is only appended during ResolveTurnAsync.
--- a/tests/Pinder.Core.Tests/Conversation/Issue293_MarkEndedTests.cs
+++ b/tests/Pinder.Core.Tests/Conversation/Issue293_MarkEndedTests.cs
@@ -1,0 +1,93 @@
+using System.Threading.Tasks;
+using Xunit;
+using Pinder.Core.Characters;
+using Pinder.Core.Conversation;
+using Pinder.Core.Interfaces;
+using Pinder.Core.Traps;
+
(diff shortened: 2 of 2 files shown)


decay256 (commented):
## Code Review (companion review for pinder-web#318)

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

### Blocking
_None._

### Non-blocking
- [`src/Pinder.Core/Conversation/GameSession.cs:223-231`] `MarkEnded` unconditionally writes `_ended` and `_outcome` with no precondition check. Idempotent on an already-ended session (silently re-stamps the outcome) and silently freezes a mid-game session if a caller misuses it. The XML doc is the only thing pointing callers at the legitimate post-game rehydration path. Consider either (a) a `Debug.Assert(!_ended)` to flag double-calls in dev builds, or (b) a one-liner in the doc-comment stating that calling this mid-game is undefined behaviour. Matches the codebase's permissive `RestoreState` style, so this is a style/discoverability nit, not a correctness issue.
- [`tests/Pinder.Core.Tests/Conversation/Issue293_MarkEndedTests.cs`] `[Theory]` covers all three `GameOutcome` values + a typed-exception integration test (good). Missing cheap-to-add edges:
  - `MarkEnded` called twice (idempotency / outcome override behaviour).
  - `MarkEnded` called on a mid-game session (locks current behaviour in even if it's "undefined").
  Production only calls this once on the rehydrate path, so low priority — but useful as a contract pin.

### Positive notes
- Method body is the absolute minimum: two field assignments. No surprises, no allocation, no side-effects beyond the documented flags.
- XML doc explicitly contrasts the API with `RestoreState` (mid-game resim) and points callers at the post-game rehydration use case. Clear contract.
- `GameOutcome` enum is small (3 values: `DateSecured`, `Unmatched`, `Ghosted`) and the theory test covers all of them — full enum-surface coverage.
- `Issue293_StartTurnAsync_ThrowsGameEndedExceptionWithOutcome` validates the end-to-end contract: after `MarkEnded(outcome)`, `StartTurnAsync` throws `GameEndedException` with the matching `Outcome`. This is the actual behaviour pinder-web's rehydrate path depends on.
- Adding a public method is non-breaking; no overload ambiguity (no other `MarkEnded` on the type), no signature collisions with existing `Restore*` / `Mark*` members.

### Test results

`Pinder.Core.Tests` filtered to `Issue293`:
```
Passed!  - Failed:     0, Passed:     5, Skipped:     0, Total:     5, Duration: 9 ms
```

**Verdict**: APPROVE

PR #299, opened 2026-04-03
[Issue #270] Write tests: Missing: 5 shadow reduction events from §7 not implemented

decay256:
Fixes #270

## DoD Evidence
**Branch:** issue-270-write-tests-missing-5-shadow-reduction-e
**Commit:** e6ad986

## Test Summary
22 spec-driven tests in `ShadowReductionSpecTests.cs` covering all 5 acceptance criteria:

| AC | Reduction | Tests |
|----|-----------|-------|
| AC-1 | Dread −1 on DateSecured | 5 tests (positive, negative delta, non-DateSecured, co-existing events, growth event text) |
| AC-2 | Denial −1 on Honesty success ≥15 | 5 tests (boundary at 15, failure, wrong stat, stacking, growth event) |
| AC-3 | Madness −1 on Recover | 5 tests (success, failure, negative delta, null shadows, Overthinking still grows) |
| AC-4 | Overthinking −1 on disadvantage success | 5 tests (success, failure, wrong stat, low threshold, growth event) |
| AC-6 | Null shadow tracker safety | 2 tests (DateSecured, Honesty) |

Each test has a mutation comment explaining what specific code change would cause it to fail.

PR #771, opened 2026-04-28
feat(#299): pass typed phase through ILlmTransport

decay256:
Adds an optional `phase` parameter to `ILlmTransport.SendAsync` and `IStreamingLlmTransport.SendStreamAsync` so decorators (snapshot recorders, telemetry) can label exchanges without re-deriving the phase from prompt text.

This is the pinder-core half of decay256/pinder-web#299.

## Changes

- New `LlmPhase` constants in `Pinder.Core.Interfaces` (canonical phase strings: `dialogue_options`, `delivery`, `steering`, `opponent_response`, `interest_change_beat`, `horniness_overlay`, `shadow_corruption`, `matchup_analysis`, `psychological_stake`, `unknown`).
- Optional `string? phase = null` parameter on `SendAsync` and `SendStreamAsync` — **backward compatible**, existing callers and implementations don't need to change.
- `AnthropicTransport` / `OpenAiTransport` / streaming counterparts accept and ignore the value (provider has no use for it).
- `PinderLlmAdapter` passes the canonical phase string at every call site.
- `LlmStakeGenerator` / `LlmMatchupAnalyzer` pass `psychological_stake` / `matchup_analysis` on both non-streaming and streaming paths.
- Test stubs (`StubLlmTransport`, `FakeStreamingTransport`) updated; `FakeStreamingTransport` now captures `LastPhase` for assertions.

## Backward compatibility

The interface change uses an optional default-null parameter, so existing implementations of `ILlmTransport` / `IStreamingLlmTransport` continue to compile and run without modification. Only test stubs that **explicitly** declare the full method signature needed updating (two of them).

## Pairs with

decay256/pinder-web PR `fix/299-typed-phase-param` — that PR bumps the submodule pointer here, swaps `SnapshotRecordingLlmTransport` to read the phase from the new param, and deletes `LlmExchangePhaseDetector`.

Merge this first.

### DIFF
--- a/src/Pinder.Core/Interfaces/ILlmTransport.cs
+++ b/src/Pinder.Core/Interfaces/ILlmTransport.cs
@@ -16,7 +16,14 @@ namespace Pinder.Core.Interfaces
         /// <param name="userMessage">The user-turn message content.</param>
         /// <param name="temperature">Sampling temperature (default 0.9).</param>
         /// <param name="maxTokens">Maximum tokens for the response (default 1024).</param>
+        /// <param name="phase">
+        /// Optional engine-phase label (see <see cref="LlmPhase"/>). Transports themselves
+        /// should ignore the value; decorators (snapshot recorders, telemetry) read it to
+        /// classify the exchange without inspecting prompt text. Defaults to <c>null</c>
+        /// for backwards compatibility — existing callers and ILlmTransport implementations
+        /// do not need to change.
+        /// </param>
         /// <returns>Raw text response from the LLM.</returns>
-        Task<string> SendAsync(string systemPrompt, string userMessage, double temperature = 0.9, int maxTokens = 1024);
+        Task<string> SendAsync(string systemPrompt, string userMessage, double temperature = 0.9, int maxTokens = 1024, string? phase = null);
     }
 }
--- a/src/Pinder.Core/Interfaces/IStreamingLlmTransport.cs
+++ b/src/Pinder.Core/Interfaces/IStreamingLlmTransport.cs
@@ -34,6 +34,12 @@ namespace Pinder.Core.Interfaces
         /// <param name="temperature">Sampling temperature (default 0.9).</param>
         /// <param name="maxTokens">Maximum tokens for the response (default 1024).</param>
         /// <param name="cancellationToken">Cancellation token; cancelling stops the stream.</param>
+        /// <param name="phase">
+        /// Optional engine-phase label (see <see cref="LlmPhase"/>). Transports themselves
+        /// should ignore the value; decorators (snapshot recorders, telemetry) read it to
+        /// classify the exchange without inspecting prompt text. Defaults to <c>null</c>
+        /// for backwards compatibility.
+        /// </param>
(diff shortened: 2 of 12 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 1

### Blocking
_(none)_

### Non-blocking

- **No test asserts that production call sites in `PinderLlmAdapter` / `LlmStakeGenerator` / `LlmMatchupAnalyzer` pass the right phase.** `FakeStreamingTransport` was extended with a `LastPhase` capture field (`tests/Pinder.Core.Tests/SessionSetup/FakeStreamingTransport.cs:26,65`) but nothing reads it. The 11 `phase: LlmPhase.X` arguments in `PinderLlmAdapter.cs` plus the 4 in `LlmStakeGenerator` / `LlmMatchupAnalyzer` could silently regress to `phase: null` (or to a wrong constant) without a single test failing. Worth at least one assertion per phase cluster — the captures are already there on the fakes, just unused.

### Positive notes

- **`LlmPhase` constants are well-defined.** All 10 canonical strings (`dialogue_options`, `delivery`, `steering`, `opponent_response`, `interest_change_beat`, `horniness_overlay`, `shadow_corruption`, `matchup_analysis`, `psychological_stake`, `unknown`) match the values the consuming snapshot recorder in pinder-web previously emitted via `LlmExchangePhaseDetector`. Continuity preserved.
- **Doc comment on `LlmPhase` correctly flags it as a public contract.** The "external decorators rely on these exact values — add new constants instead of renaming" note is the right warning.
- **Backward-compat is real, not theoretical.** `string? phase = null` is appended to both `ILlmTransport.SendAsync` and `IStreamingLlmTransport.SendStreamAsync`. For streaming, `phase` sits *after* `CancellationToken cancellationToken = default`, so existing positional callers keep working. Existing implementations of either interface don't need to recompile — verified by the `_ = phase;` discard pattern in `AnthropicTransport`, `AnthropicStreamingTransport`, `OpenAiTransport`, `OpenAiStreamingTransport`.
- **Streaming call sites use named `phase:` arguments.** `LlmStakeGenerator.cs:103` and `LlmMatchupAnalyzer.cs:166` both name the argument explicitly, which is robust against future signature reordering.
- **Production transports correctly drop the value.** Each adapter has a `// phase is metadata for decorators; the underlying provider has no use for it.` comment plus `_ = phase;` — clear intent, no risk of accidentally including `phase` in an outbound LLM API payload.
- **All 11 `PinderLlmAdapter` call sites pass an explicit phase.** Including the three positional opponent-response variants in `SendStatefulOpponentAsync` (empty history / single message / multi-turn) — none silently default to null.
- **Tests run green.** SessionSetup subset: 23/23 passing.

### Test results

```
tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj --filter "FullyQualifiedName~SessionSetup"
Passed!  - Failed: 0, Passed: 23, Skipped: 0, Total: 23, Duration: 104ms
```

**Verdict**: APPROVE

PR #320, opened 2026-04-03
[Issue #314] Spec: XP risk-tier multiplier (1x/1.5x/2x/3x)

decay256:
Refs #314

## Summary
Specification document for the missing XP risk-tier multiplier. Documents the multiplier logic (Safe=1x, Medium=1.5x, Hard=2x, Bold=3x) to be applied to base XP on successful checks in `GameSession.RecordRollXp`.

## Spec File
`docs/specs/issue-314-spec.md`

## DoD Evidence
**Branch:** issue-314-write-spec-document-missing-xp-risk-tier
**Commit:** a661f44
**Spec covers:** function signatures, I/O examples, all 6 acceptance criteria, edge cases (Nat20/Nat1/failure bypass, rounding), error conditions, dependencies

Issue #348, opened 2026-04-03
Player agent: LlmPlayerAgent — Sonnet/Opus plays the game with full rules context


decay256:
## Context
An LLM-backed player agent that reads the rules, understands the current game state, and makes strategic choices with written reasoning. Produces the most realistic simulation of a thoughtful human player.

## What it does
1. Receives all 4 options with full mechanical context (DC, %, risk tier, hidden bonuses, DCs adjusted for momentum/tells)
2. Receives current game state (interest, momentum, shadow values, traps, turn number)
3. Has the full rules reference available (or a condensed version)
4. Returns: chosen option index + written explanation of the trade-off reasoning

## LLM prompt structure
```
You are playing Pinder, a comedy dating RPG. You are {character_name}, a sentient penis on a dating app.
You must choose one of 4 dialogue options. Your goal is to reach Interest 25 (date secured).

## Current State
- Interest: {N}/25 ({state}) — {modifier note}
- Momentum: {streak} wins {bonus note}
- Active traps: {list or 'none'}
- Your shadow levels: {shadow states}
- Turn: {N}

## Your Options
A) [{STAT} +{mod}] DC {dc} | {need}+ | {pct}% | {risk_tier} {icons}
   Intended: "{text}"
   Expected interest: +{expected} | XP: {xp}
   Bonuses: {list}

[B, C, D same format]

## Rules reminder
- Safety: beat DC on d20+mod+bonuses ≥ DC
- Miss 1-2: Fumble (−1 interest). Miss 3-5: Misfire (−1). Miss 6-9: Trope Trap (−2 + 1-2 turn trap on this stat).
- Miss 10+: Catastrophe (−3 + trap). Nat 1: Legendary fail (−4).
- Hard success: +1 bonus interest. Bold: +2 bonus interest.
- Momentum: 3 wins in a row → +2 to next roll.
- Callback icons 🔗 indicate a hidden +1/+2/+3 bonus to the roll.
- Tell icon 📖 = +2 hidden bonus. Combo icon ⭐ = +1 interest on success.

Choose the option you think gives the best outcome. Explain your trade-off thinking first, then state your choice as: PICK: [A/B/C/D]
```

## Implementation
- Model: Sonnet (default) or configurable via env var
- Uses `ANTHROPIC_API_KEY` — same as `AnthropicLlmAdapter`
- Returns `PlayerDecision` with LLM reasoning in `Reasoning` field
- Falls back to `ScoringPlayerAgent` if API fails

[...]

PR #772, opened 2026-04-28
fix(transports): surface assistant text from thinking/reasoning channels

decay256:
Fixes the parser-level part of pinder-web #320.

## Problem

When the engine is configured with a thinking-capable model (Anthropic Claude with extended thinking, OpenAI o-series, gpt-5* with reasoning enabled, or any provider routing through OpenRouter that emits reasoning frames), the non-streaming transports silently returned the empty string and downstream callers (LlmMatchupAnalyzer, LlmStakeGenerator, AnthropicLlmAdapter delivery / opponent fallback) reported `matchup_llm_failed: matchup stage produced empty output`.

### 1. Anthropic native thinking blocks
`MessagesResponse.GetText()` returned `Content[0].Text`. Anthropic extended-thinking responses lead with a `type:"thinking"` block followed by the real `type:"text"` answer, so `GetText()` returned "" (the thinking block carries reasoning under a different field, not `text`). Every consumer of `GetText()` saw empty strings.

**Fix:** skip every non-`text` block (`thinking`, `redacted_thinking`, `tool_use` — the latter is exposed via `GetToolInput()`) and concatenate every `text` block in order.

### 2. OpenRouter reasoning channels (non-streaming)
`OpenAiClient.SendChatCompletionAsync` only read `choices[0].message.content`. OpenRouter (the routing layer for `openrouter/anthropic/claude-opus-4.7` and any reasoning-capable model) puts the user-visible answer in `message.reasoning` and/or `message.reasoning_details[i].summary` when `content` is empty/null.

The streaming transport already handled this in pinder-core #178/#764; the non-streaming path was the missing twin.

**Fix:** extract in priority order `content` → `reasoning` → concatenated `reasoning_details` summaries (mirrors `OpenAiStreamingTransport.ExtractContentFragmentsOrThrow`).

## Regression coverage

* `MessagesResponseTests` — new tests for thinking/redacted_thinking/tool_use skipping, concatenation across multiple text blocks, thinking-only payload returns empty string.
* `SpecDtoTests.MessagesResponse_GetText_ConcatenatesAllTextBlocks` — the previous AC2 test asserted "first block only" and was enforcing the bug; rewritten to assert concatenation.
* `OpenAiClientReasoningTests` — 9 new scenarios covering the `content` → `reasoning` → `reasoning_details` fallback ladder, including whitespace-only content, missing/empty choices, and content-wins precedence.

## Test results

`Pinder.LlmAdapters.Tests` (the affected assembly):

| | Failed | Passed | Total |
|---|---|---|---|
| `origin/main` baseline | 63 | 886 | 958 |
| this branch              | 63 | 898 | 970 |
| **delta**                | **0** | **+12** | **+12** |

12 new tests, all pass; zero pre-existing tests regressed. The 63 pre-existing failures on main are unrelated dialogue-options spec tests (`AnthropicLlmAdapterIssue208Tests`, `Issue240_DialogueOptionsFormatTests`, etc.) that fail on the bare `main` SHA `be19c29` and are out of scope for this PR.

`Pinder.GameApi.Tests` (downstream consumer of these DTOs): 382 / 382 pass.

## Out of scope

* Inline `<thinking>...</thinking>` tag stripping (some non-native-thinking models emit these in user-visible content). The structured paths in this code base use Anthropic `tool_use` (handled by `GetToolInput`) or prose (matchup / stakes), so inline tags would only manifest as cosmetic noise rather than parse failure. Filing as a follow-up if it becomes user-visible.
* OpenAI `reasoning_content` (Chat Completions `o1`-style field — distinct from OpenRouter's `reasoning`). If/when we wire o1-* directly without OpenRouter we should add it to the same ladder.

Closes the parser portion of decay256/pinder-web#320 (the pinder-web bump PR carries the issue closer).

### DIFF
--- a/src/Pinder.LlmAdapters/Anthropic/Dto/MessagesResponse.cs
+++ b/src/Pinder.LlmAdapters/Anthropic/Dto/MessagesResponse.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Text;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -16,9 +17,41 @@ namespace Pinder.LlmAdapters.Anthropic.Dto
         public UsageStats? Usage { get; set; }
 
         /// <summary>
-        /// Returns the text of the first content block, or empty string if no content blocks exist.
+        /// Returns the user-visible assistant text from the response.
+        /// Concatenates every <c>type: "text"</c> content block in order and skips
+        /// non-text blocks such as <c>thinking</c> / <c>redacted_thinking</c>
+        /// (Anthropic extended thinking — issue #320) and <c>tool_use</c> (which is
+        /// surfaced separately via <see cref="GetToolInput"/>). Returns the empty
+        /// string when no text blocks are present.
         /// </summary>
-        public string GetText() => Content.Length > 0 ? Content[0].Text ?? "" : "";
+        /// <remarks>
+        /// Pre-#320 this method returned <c>Content[0].Text</c>. That was correct
+        /// only when the model emitted text as the very first block; with extended
+        /// thinking enabled the first block is <c>type: "thinking"</c> and the
+        /// real answer sits in a later <c>type: "text"</c> block, so the old
+        /// behaviour returned the empty string and downstream parsers reported
+        /// <c>matchup_llm_failed</c> / similar empty-output errors.
+        /// </remarks>
+        public string GetText()
+        {
+            if (Content == null || Content.Length == 0) return "";
+            StringBuilder? sb = null;
+            for (int i = 0; i < Content.Length; i++)
+            {
+                var block = Content[i];
+                if (block == null) continue;
+                // Only "text" blocks contribute to the user-visible answer. Thinking
+                // and redacted_thinking blocks are deliberately suppressed; tool_use
+                // is exposed separately via GetToolInput().
+                if (!string.Equals(block.Type, "text", StringComparison.OrdinalIgnoreCase))
+                    continue;
+                var text = block.Text;
+                if (string.IsNullOrEmpty(text)) continue;
+                if (sb == null) sb = new StringBuilder(text.Length);
+                sb.Append(text);
+            }
+            return sb?.ToString() ?? "";
+        }
 
         /// <summary>
         /// Returns the tool input from the first tool_use content block, or null if none.
--- a/src/Pinder.LlmAdapters/OpenAi/OpenAiClient.cs
+++ b/src/Pinder.LlmAdapters/OpenAi/OpenAiClient.cs
@@ -74,8 +74,13 @@ namespace Pinder.LlmAdapters.OpenAi
                         try
                         {
                             var json = JObject.Parse(body);
-                            var text = json["choices"]?[0]?["message"]?["content"]?.ToString();
-                            return text ?? "";
+                            return ExtractAssistantText(json);
+                        }
+                        catch (InvalidOperationException)
+                        {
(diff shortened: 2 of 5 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

**Verdict: APPROVE**

Reviewed in tandem with the pinder-web PR #348 that bumps to this commit. Findings specific to pinder-core:

### Verification

**`MessagesResponse.GetText()` — concat + non-text-block skip**
- Skips `thinking`, `redacted_thinking`, `tool_use` (case-insensitive); concatenates every `text` block in source order.
- Defensive against `Content == null`, individual null blocks, null/empty `block.Text`.
- Empty text-block list returns `""` (not null, not exception) — pinned.
- 5 new tests in `MessagesResponseTests` + 1 rewrite in `SpecDtoTests` (the previous "first block only" test was enforcing the bug — now corrected).

**`OpenAiClient.ExtractAssistantText` — reasoning ladder**
- Order: `content` → `reasoning` → concatenated `reasoning_details[i].summary`. Matches the source-order policy of `OpenAiStreamingTransport.ExtractContentFragmentsOrThrow` but with **first-non-empty** semantics rather than emit-all (a deliberate divergence — non-streaming consumers parse the result as structured prose and reasoning leakage would corrupt that).
- Empty-`choices` handled with explicit `Count == 0` guard, not `?[0]?` (which doesn't behave like a Dictionary lookup on `JArray`).
- `message.Type != JTokenType.Object` guards malformed shapes.
- Whitespace-only content correctly falls through to reasoning.
- 9 new tests cover the full ladder and edge cases.

### Test results (fresh worktree at `bed3e92`)

```
Pinder.LlmAdapters.Tests:  898 passed / 63 failed / 9 skipped / 970 total
```

Verified the 63 failures match `origin/main` (`be19c29`) baseline exactly — they are pre-existing dialogue-options spec failures unrelated to this PR. **+12 new tests, all pass; 0 regressions.**

### Non-blocking

1. **Doc comment overstates streaming parity.** `ExtractAssistantText`'s `<remarks>` says "mirrors `OpenAiStreamingTransport.ExtractContentFragmentsOrThrow`", but the streaming variant emits all populated channels per frame while this method returns the first non-empty channel only. Suggest clarifying: "Uses the same source order as `ExtractContentFragmentsOrThrow`, but returns the first non-empty channel rather than concatenating across channels (avoids reasoning leakage into structured-prose consumers)."

2. **Dead `catch (InvalidOperationException)` block** in `SendChatCompletionAsync`. `ExtractAssistantText` only throws `ArgumentNullException`, never `InvalidOperationException`. The catch comment claims `ExtractAssistantText` "raises a typed error for malformed shapes" but malformed shapes actually return `""`. Either remove the catch or extend `ExtractAssistantText` to throw `InvalidOperationException` for genuinely malformed payloads (e.g., `choices` present but not a `JArray`, or `message` present but not an object). Cosmetic — current code is functionally safe.

### Positive notes

- Tests use real OpenRouter `reasoning_details` shape (`{type: "reasoning.summary", summary: "..."}`), not synthetic. Same goes for the Anthropic `thinking` / `redacted_thinking` block fixtures.
- The `SpecDtoTests` rewrite correctly identifies and replaces the test that was previously codifying the bug.
- Streaming-vs-non-streaming policy divergence is the right call (preventing reasoning leakage into JSON parsers downstream).

---
Reviewed at `bed3e92`. Coordinated with pinder-web #348.

PR #340, opened 2026-04-03
[Issue #314] Apply XP risk-tier multiplier (1x/1.5x/2x/3x)

decay256:
Fixes #314

## What was changed
Applied risk-tier XP multiplier to base success XP in `GameSession.RecordRollXp`:
- Safe: 1x (no change)
- Medium: 1.5x (rounded to nearest int)
- Hard: 2x
- Bold: 3x

Nat20 (25 XP), Nat1 (10 XP), and Failure (2 XP) are NOT affected by the multiplier — they use flat values per §10 precedence rules.

## Implementation
Added `ApplyRiskTierMultiplier(int baseXp, RiskTier riskTier)` static method to GameSession. Uses if/else chain (C# 8.0 compatible, no switch expressions with pattern matching).

## How to test
```bash
dotnet test --filter "XpTrackingSpecTests|XpTrackingGameSessionTests"
```

## Tests
- 9 new dedicated risk-tier multiplier tests (Safe/Medium/Hard/Bold, Nat20/Nat1/Failure no-multiplier, rounding)
- Updated 15 existing XP tests to expect multiplied values
- All 1931 tests pass (1447 Core + 484 LlmAdapters)

## Deviations from contract
None

## DoD Evidence
**Branch:** issue-314-missing-xp-risk-tier-multiplier-1x-1-5x-
**Commit:** c23cc3e

PR #339, opened 2026-04-03
[Issue #313] Write tests: Missing: Lukewarm (5-9) not a distinct InterestState — rules §6 defines it explicitly

decay256:
Fixes #313

## DoD Evidence
**Branch:** issue-313-write-tests-missing-lukewarm-5-9-not-a-d
**Commit:** 95e4f76

## Test Summary
55 tests covering all acceptance criteria:
- AC1: `InterestState.Lukewarm` enum exists with correct ordinal position
- AC2: `GetState()` returns Lukewarm for values 5-9 (all 5 values)
- AC3: `GetState()` returns Interested for 10-15 (all 6 values)
- AC4: Boundary transitions (Bored↔Lukewarm, Lukewarm↔Interested)
- AC5: `GrantsAdvantage`/`GrantsDisadvantage` unaffected by Lukewarm
- AC6: All other states unaffected by enum insertion
- AC7: Full boundary sweep (0-25 → correct states)
- Default starting value (10) is Interested, not Lukewarm

All tests annotated with mutation comments.

PR #333, opened 2026-04-03
[Issue #310] Write tests: Missing: Madness T3 (≥18) should replace one dialogue option with unhinged text

decay256:
Fixes #310

## DoD Evidence
**Branch:** issue-310-write-tests-missing-madness-t3-18-should
**Commit:** e216680

### AC1: DialogueOption.IsUnhingedReplacement property
- Default is false
- Can be constructed with true
- Explicit false stays false

### AC2: Madness T3 (≥18) marks exactly one option
- Exactly 18 → one option marked
- 30 (far above) → still exactly one
- Preserves original Stat
- Preserves original IntendedText
- Only one option marked, not all

### AC3: Below T3, no options marked
- Madness 12 (T2) → none marked
- Madness 17 (boundary) → none marked
- Madness 0 → none marked
- Null shadow tracker → none marked

### AC5: Edge cases
- Single option + Madness T3 → that option marked
- Empty options + Madness T3 → no crash

### Edge cases
- Fixation T3 + Madness T3 → both apply

Issue #350, opened 2026-04-03
Session runner: enable shadow tracking by passing GameSessionConfig with PlayerShadows


decay256:
## Context
The session runner creates `GameSession` without a `GameSessionConfig`, meaning `_playerShadows` is null and all shadow growth triggers silently do nothing. The rules intend shadows to grow during play — in a simulation this is critical for testing the shadow mechanic.

## Fix
Create a `SessionShadowTracker` for the player, pass it via `GameSessionConfig`, and display shadow state changes in the playtest output.

```csharp
// In session-runner/Program.cs
var sableShadows = new SessionShadowTracker(
    new Dictionary<ShadowStatType, int> {
        { ShadowStatType.Denial, 3 },    // from character profile
        { ShadowStatType.Fixation, 2 },
        // others start at 0
    });

var config = new GameSessionConfig {
    PlayerShadows = sableShadows,
    // OpponentShadows can also be set for full simulation
};

var session = new GameSession(sable, brick, llm, dice, traps, config);
```

## Playtest output enhancement
When shadow grows, show it in the session turn output:
```
⚠️ SHADOW GROWTH: Fixation +1 (same stat 3 turns in a row) → Fixation now 3
```

At session end, show final shadow state delta:
```
## Shadow Changes This Session
| Shadow | Start | End | Delta |
|---|---|---|---|
| Denial | 3 | 4 | +1 (date without Honesty success) |
| Fixation | 2 | 2 | 0 |
```

## Acceptance Criteria
- [ ] Session runner passes `GameSessionConfig` with `PlayerShadows` from character's starting shadow values
- [ ] Shadow growth events appear in turn output when they fire
- [ ] Session summary includes shadow delta table
- [ ] Test: running a session where same stat is picked 3 turns in a row shows Fixation +1 in output
- [ ] Build clean

Depends on: #346 (needs player agent to vary stat picks so shadow triggers actually fire)

Depends on: #346

decay256:
## Work Summary — Code Review (Issue #350)

**Agent**: code-reviewer
**Correlation ID**: doc-review-sprint-decay256-pinder-core-1-1775257091167-issue-350-r1
**PR**: #393

### Review Result: REQUEST_CHANGES

**1 Error found:**
- `issue-348-spec.md` uses `Obsession` as a `ShadowStatType` enum value, but the actual enum defines `Overthinking`. This factual error will mislead the implementer.

**1 Warning found:**
[...]

PR #773, opened 2026-04-28
feat: post-LLM text-processing layer (#340 em-dash, #339 callback strip, #333 turn-0 scene)

decay256:
## Summary

Three engine-side text-processing changes, packaged together because they all touch the post-LLM text-handling layer in pinder-core.

### #340 — em-dash → semicolon post-process
`PunctuationNormalizingTransport` decorator wraps any `ILlmTransport` (and optionally `IStreamingLlmTransport`). On every response it rewrites ` — ` → `; ` (covers ASCII space + U+2014, U+2009 thin-space, and mixed variants). `word—word` and en-dashes (U+2013) are intentionally preserved.

The web tier wraps the provider transport in this decorator BEFORE the snapshot recorder so audit logs capture the post-normalised text.

### #339 — same-turn callback strip
`CallbackStripper` regex pass that runs in `GameSession.ResolveTurnAsync` after the existing Steering / Horniness / Shadow layers. Strips conservative same-turn callback patterns ("As you (just) said,", "Like we mentioned,", "As we discussed,", etc.) and emits a `Callback Strip` `TextDiff` layer when it changes the message — same diff plumbing as the other text-transform layers.

Adds `LlmPhase.CallbackStrip` and `LlmPhase.OutfitDescription` (latter consumed by #333).

Explicit cross-turn references ("earlier on turn 2…") short-circuit the strip and are returned unchanged.

### #333 — turn-0 bios + outfit description
- New `IOutfitDescriber` + default `LlmOutfitDescriber` (sibling to `IMatchupAnalyzer` / `IStakeGenerator`). Single non-streaming call, returns empty string on transport failure.
- New `Senders.Scene` constant (`"[scene]"`) for synthetic scene-entry sender tag.
- `GameSession.SeedSceneEntries(playerBio, opponentBio, outfitDescription)` appends up to three `[scene]` entries to the conversation log; throws if any turn has resolved.
- New private `BuildHistoryForLlmContext()` filter excludes `[scene]` entries from the history view fed to subsequent LLM calls so the analyzer never sees its own scene-description output.
- `TurnSnapshot.ConversationEntry` comment block updated per AGENTS.md schema discipline.

The pinder-web tier wires this up in `ActiveSession.SetupAsync`, running the outfit call in parallel with matchup analysis (`Task.Run` + join after stakes).

## Tests

| Layer | Added | Notes |
|---|---|---|
| #340 | 13 | Pure normalisation + decorator behaviour, both streaming and non-streaming |
| #339 | 13 | Stripped openers, mid-sentence cleanup, capitalisation, cross-turn preservation |
| #333 | 5 | Seeded entries, empty-skip, post-turn guard, LLM-context filter, ResimulateData round-trip |

```
Passed!  - Pinder.Core.Tests   — 30/30 new tests pass
Passed!  - Pinder.LlmAdapters.Tests — 13/13 new tests pass
```

Pre-existing baseline failures (`CharacterLoaderSpecTests` — design/examples directory missing, YamlDotNet test infrastructure bug) confirmed unchanged on baseline by stash + re-test.

## Companion PR
pinder-web: https://github.com/decay256/pinder-web/pull/new/fix/A1-engine-text-and-turn0 — bumps the submodule pointer + wires up the LlmOutfitDescriber + ICharacterRepository.Items.

**Merge order: this PR first, then the pinder-web PR.**

### DIFF
--- a/session-runner/Snapshot/SessionSnapshot.cs
+++ b/session-runner/Snapshot/SessionSnapshot.cs
@@ -63,7 +63,15 @@ namespace Pinder.SessionRunner.Snapshot
     ///   <item><description><c>CharmUsageCount</c>, <c>CharmMadnessTriggered</c></description></item>
     ///   <item><description><c>SaUsageCount</c>, <c>SaOverthinkingTriggered</c></description></item>
     ///   <item><description><c>RizzCumulativeFailureCount</c></description></item>
-    ///   <item><description><c>ConversationHistory</c> (per-entry sender + text + per-layer text_diffs[] — issue #305)</description></item>
+    ///   <item><description>
+    ///     <c>ConversationHistory</c> (per-entry sender + text + per-layer text_diffs[]
+    ///     — issue #305). Includes the turn-0 scene-setting entries seeded by
+    ///     <c>GameSession.SeedSceneEntries</c> (issue #333): three entries with
+    ///     sender == <c>"[scene]"</c> for player bio, opponent bio, and the
+    ///     LLM-generated outfit description. The <c>callback_strip</c> layer
+    ///     (issue #339) appears as one more entry in <c>TextDiffs</c> when
+    ///     same-turn callback phrases were stripped from the delivered message.
+    ///   </description></item>
     /// </list>
     /// </summary>
     public sealed class TurnSnapshot
--- a/src/Pinder.Core/Conversation/ConversationIndexing.cs
+++ b/src/Pinder.Core/Conversation/ConversationIndexing.cs
@@ -1,0 +1,176 @@
+using System.Collections.Generic;
+
+namespace Pinder.Core.Conversation
+{
+    /// <summary>
+    /// Helpers for mapping a physical <c>ConversationHistory</c> index
+    /// to a logical (turn-number, role) pair while transparently skipping
+    /// non-conversational <see cref="Senders.Scene"/> entries seeded at
+    /// the front of the history (issue #333).
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Pre-#333 the conversation log was strict alternating
+    /// <c>(player, opponent, player, opponent, ...)</c> pairs, so consumers
+    /// could derive the turn number with <c>(i / 2) + 1</c> and the role
+    /// with <c>i % 2 == 0</c>. Once the engine started seeding
+    /// <c>[scene]</c> entries (player bio, opponent bio, outfit
+    /// description) at indices 0..N before the first turn, that pair-math
+    /// shifts by N and silently misattributes turn numbers / text-diffs.
+    /// </para>
+    /// <para>
+    /// This class encapsulates the "skip scenes, then pair-math the
+    /// remaining entries" rule so every consumer agrees on the same
+    /// mapping. Two patterns are exposed:
+    /// </para>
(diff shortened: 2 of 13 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

Engine-side review for #340 + #339 + #333. Tests run green: `Pinder.Core.Tests` (#333+#339) 30/30, `Pinder.LlmAdapters.Tests` (#340) 13/13.

### Non-blocking

- [src/Pinder.Core/Text/CallbackStripper.cs:80-90] `PreservePatterns` short-circuits the entire message if any `\bturn\s+\d+\b` reference is found anywhere in the input. Conservative in the right direction (false-negative beats false-positive on stylistic stripping), but it also means a message like *"As you said, that was wild. Earlier on turn 3 we…"* is left fully unstripped because of the late cross-turn reference. Probably fine for P3 but worth a one-line comment so future maintainers know the all-or-nothing semantics are intentional.

- [src/Pinder.LlmAdapters/PunctuationNormalizingTransport.cs:127-130] `IsConvertibleSpace` accepts ASCII space (U+0020) and U+2009 thin-space. Some providers occasionally emit U+00A0 (non-breaking space) around em-dashes; not in scope for the issue spec, but if you see ` ` em-dash patterns surviving in audit logs in the field, that's the next variant to add.

### Positive notes

- `Senders.IsScene` + `BuildHistoryForLlmContext()` filter is wired into all three LLM-history sites in `GameSession.ResolveTurnAsync` (lines 475, 789, 985) plus the steering call (836). Scene entries cannot leak back into matchup / delivery / opponent / steering as conversation context. Hot-path branch avoids the copy when no scene entries exist.
- `SeedSceneEntries` correctly throws `InvalidOperationException` if any turn has resolved — protects against late seeding regressions.
- `LlmPhase.OutfitDescription` and `LlmPhase.CallbackStrip` constants are documented (the latter explicitly notes "no LLM call — purely a regex post-process" so future readers don't expect a transport invocation).
- `CallbackStripper` runs AFTER Steering / Horniness / Shadow in `GameSession` (line 949), matching the PR's stated intent — operates on the final delivered text before it is committed to `_history`. The `WordDiff.Compute(beforeCallbackStrip, deliveredMessage)` argument order is correct (before→after).
- `LlmOutfitDescriber` mirrors `IStakeGenerator`'s catch-all → empty-string contract; transport failure does not fail setup, just elides one scene entry. Defensive and consistent with siblings.
- `TurnSnapshot.ConversationEntry` comment block in `session-runner/Snapshot/SessionSnapshot.cs` is updated for both #333 (scene entries) and #339 (callback_strip layer) per AGENTS.md schema discipline.

### Note on the companion web PR

#350 (pinder-web) has two blocking issues stemming from this PR's introduction of `[scene]` entries: `GET /sessions/{id}/log` and the persistence loop in `ActiveSession` both make a `pair-math` assumption (`Turn = i/2 + 1`, `isPlayerEntry = hi % 2 == 0`) that is now broken with non-conversational entries at the front of the history. The engine-side change is correct on its own — those are consumer-side regressions. Flagging here since the cross-repo merge order in the web PR body says #773 lands first.

**Verdict: APPROVE** — engine work is sound; flagged consumer-side regressions live in #350.

PR #332, opened 2026-04-03
[Issue #310] Madness T3 (≥18) replaces one dialogue option with unhinged marker

decay256:
Fixes #310

## What was implemented

- Added `IsUnhingedReplacement` bool property to `DialogueOption` (default `false`, backward-compatible)
- Added Madness T3 block in `GameSession.StartTurnAsync()` after Denial T3: when Madness raw value ≥ 18, one random option is marked `IsUnhingedReplacement = true`
- Fixed Horniness T3 block to preserve `IsUnhingedReplacement` when converting options to Rizz stat

## How to test

```bash
dotnet test --filter ShadowThresholdGameSession
```

7 new tests:
- `MadnessT3_MarksExactlyOneOptionAsUnhinged` — T3 marks exactly one option
- `MadnessT2_NoOptionsMarkedUnhinged` — T2 marks none
- `MadnessT3_PreservesOriginalStatAndText` — unhinged option keeps original stat/text
- `MadnessT3_WithHorninessT3_PreservesUnhingedFlag` — flag survives Horniness T3
- `MadnessBelow18_NoUnhingedOptions` — value 17 marks none
- `DialogueOption_IsUnhingedReplacement_DefaultsFalse` — backward compat

All 1799 tests pass (1346 Core + 453 LlmAdapters).

## Deviations from contract
None

## DoD Evidence
**Branch:** issue-310-missing-madness-t3-18-should-replace-one
**Commit:** 45c31ef

Issue #382, opened 2026-04-03
Test coverage: add SuccessScaleTests.cs with all tier boundaries


decay256:
## Context
`SuccessScale.GetInterestDelta()` has 4 tiers:
- Beat by 1-4 → +1 Interest
- Beat by 5-9 → +2 Interest
- Beat by 10+ → +3 Interest (Crit)
- Nat 20 → +4 Interest

These are tested indirectly through GameSession integration tests but have no direct unit tests.

## What to create
`tests/Pinder.Core.Tests/SuccessScaleTests.cs` with a Theory covering all tiers and boundaries:
- margin=1 → +1
- margin=4 → +1 (boundary)
- margin=5 → +2
- margin=9 → +2 (boundary)
- margin=10 → +3
- margin=50 → +3 (large margin still +3)
- Nat 20 → +4 (regardless of margin)
- margin=0 (failure) → 0

## Acceptance Criteria
- [ ] `SuccessScaleTests.cs` created with all tier boundaries as Theory
- [ ] Tests are direct unit tests against `SuccessScale` class, not through GameSession
- [ ] Build clean, all tests pass

**Role**: test-engineer
**Maturity**: prototype

decay256:
Closing: test cleanup/coverage issue — superseded by the comprehensive test suite added across recent sprints (2200+ tests). If specific gaps surface during development, file a targeted issue.

PR #776, opened 2026-04-29
feat(SessionSetup): add LlmMatchupSummarizer for #332

decay256:
Adds a new `IMatchupSummarizer` + `LlmMatchupSummarizer` pair under `Pinder.SessionSetup` to produce a short 1-2 paragraph plain-prose summary of a matchup, intended as a display-only affordance alongside the existing long-form matchup analysis.

## Why a separate analyzer (and not extending `LlmMatchupAnalyzer`)?

The recommended approach in the issue was \"extend the matchup-analysis prompt to produce both outputs in one call\". Considered, rejected for these reasons:

1. The existing matchup-analysis path is **streaming-first**: `IMatchupAnalyzer.StreamMatchupAsync` feeds raw token deltas into the UI as they arrive (see `SetupStreamEvent.Delta` in pinder-web). Tacking a `Summary:` section onto the end of that stream would leak the new affordance into the main matchup-analysis display.
2. `ILlmTransport.SendAsync` returns plain `string`. Structured output (Anthropic `tool_use` / OpenAI structured outputs) would require an interface change with knock-on effects through every transport implementation.
3. A separate, short, non-streaming call is parallelisable with stake generation in the web tier, so the end-to-end latency cost is roughly free.

## What's here

- `Pinder.Core.Interfaces.LlmPhase.MatchupSummary` (`\"matchup_summary\"`) — new phase string for snapshot/audit decorators.
- `Pinder.SessionSetup.IMatchupSummarizer` — narrow interface, one method, returns a string.
- `Pinder.SessionSetup.LlmMatchupSummarizer` — default implementation built on `ILlmTransport` (non-streaming). 200-token cap, T=0.7. Failure semantics mirror `IOutfitDescriber`: empty string on transport failure (the summary is non-critical).

## Constraint

The summary is **display-only**. It is NOT used as input to any subsequent LLM prompt in the session — that is the explicit hard constraint from the issue's spec. Verifiable by grep: nothing reads `_matchupSummary` / `MatchupSummary` other than the API DTO and the persistence envelope.

## Companion PR

Wired into pinder-web in [decay256/pinder-web#332](https://github.com/decay256/pinder-web/issues/332) (`Closes #332`).

## DoD evidence

`dotnet build src/Pinder.SessionSetup/Pinder.SessionSetup.csproj` clean (84 unrelated warnings, 0 errors).

### DIFF
--- a/src/Pinder.Core/Interfaces/LlmPhase.cs
+++ b/src/Pinder.Core/Interfaces/LlmPhase.cs
@@ -38,6 +38,14 @@ namespace Pinder.Core.Interfaces
         /// <summary>Session-setup matchup analysis.</summary>
         public const string MatchupAnalysis = "matchup_analysis";
 
+        /// <summary>
+        /// Session-setup short matchup summary (issue #332). A 1-2 paragraph
+        /// LLM-generated condensation of the long-form matchup analysis,
+        /// surfaced to the UI as a display-only affordance. NOT used as
+        /// input to any subsequent LLM prompt in the session.
+        /// </summary>
+        public const string MatchupSummary = "matchup_summary";
+
         /// <summary>Session-setup psychological-stake generation.</summary>
         public const string PsychologicalStake = "psychological_stake";
 
--- a/src/Pinder.SessionSetup/IMatchupSummarizer.cs
+++ b/src/Pinder.SessionSetup/IMatchupSummarizer.cs
@@ -1,0 +1,47 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Pinder.Core.Characters;
+
+namespace Pinder.SessionSetup
+{
+    /// <summary>
+    /// Issue #332: produces a short (1-2 paragraph), human-readable summary
+    /// of a matchup. Display-only — the result is surfaced to the UI in the
+    /// Player Sheet "Background" tab alongside the full matchup analysis,
+    /// and is NOT used as input to any subsequent LLM prompt in the session.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Output contract (mirrors <see cref="IMatchupAnalyzer"/> and
+    /// <see cref="IOutfitDescriber"/>): plain prose, paragraph breaks
+    /// only — no markdown, no bullets, no headings, no code formatting.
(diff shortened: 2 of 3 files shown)


decay256 (commented):
## Code Review (paired with pinder-web #382)

**Verdict**: APPROVE

This PR adds `IMatchupSummarizer` + `LlmMatchupSummarizer` + `LlmPhase.MatchupSummary` to pinder-core. Reviewed jointly with the consumer wiring in [decay256/pinder-web#382](https://github.com/decay256/pinder-web/pull/382#pullrequestreview).

### Verified
- Separate-call design rationale is sound: the existing matchup-analysis path is streaming-first, and `ILlmTransport.SendAsync` returns a single `string` — extending it for dual-output would require a structured-output interface change with knock-on effects through every transport.
- Failure semantics: catch-all returns `string.Empty`, mirroring `IOutfitDescriber` (#333). The web tier tolerates empty → leaves `Summary` null. Setup never fails over a summary error.
- Phase label `LlmPhase.MatchupSummary = "matchup_summary"` lower-snake-case, consistent with existing constants. Docstring explicitly notes "NOT used as input to any subsequent LLM prompt" — encodes the spec in the type system.
- System prompt forbids markdown (no `#`, `**`, lists, blockquotes, code) — defence-in-depth before the web tier's `MarkdownSanitizer` pass.
- Build clean: `dotnet build pinder-core/Pinder.Core.sln` → 160 unrelated warnings, 0 errors.

### Cross-repo
- pinder-web PR #382 has the submodule pointer at `1fd2657` (this PR's head). Merge order documented in PR description (pinder-core first, then pinder-web).

### Non-blocking
- The summarizer iterates `StatType` and `ShadowStatType` via inline arrays in `AppendCharacterData`. If those enums grow, the prompt will silently miss new stats. Could be replaced with a shared list constant or `Enum.GetValues<T>()` later, but matches the patterns in adjacent setup classes — not blocking.

### Positive notes
- Tight, self-contained: 3 new files, no edits to shared interfaces (`ILlmTransport` etc).
- `Options` class with sensible defaults (T=0.7, MaxTokens=200) and tunable knobs.
- Argument-null guards on both `transport` and the two `CharacterProfile` parameters.

**Verdict**: APPROVE

PR #784, opened 2026-04-30
fix(#405): floor shadow stats at 0 (read + write + restore clamps)

decay256:
## Summary

Floor shadow stats at 0 — fixes the silent state corruption documented in `decay256/pinder-web#405`. Pre-fix, stacked `-1` reductions on a low-base shadow could drive the effective value below 0; `GetEffectiveStat` then divided by 3 (C# integer division floors negatives), producing a negative penalty that **buffed the paired positive stat**.

Fix is enforced at three layers (read, write, restore) for defense in depth:

1. **Read-time clamp** in `GetEffectiveShadow` — `Math.Max(0, base + delta)`.
2. **Read-time clamp** in `GetEffectiveStat` — floor the shadow term at 0 before dividing.
3. **Write-time clamp** in `ApplyOffset` — when a negative delta would drive effective below 0, only apply down to 0; record the suppressed portion as a `(floored)` event.
4. **Restore-time clamp** in `RestoreFromSnapshot` — old/corrupt snapshots with negative target values are normalised to 0.

Audit-log convention picked: a fully-suppressed reduction reads `<Shadow> +0 (<reason>) (floored)`. A real reduction (within the floor) is unchanged. Flagging the convention here for review — alternatives considered:

- `<Shadow> -0 (<reason>) (floored)` (mirrors the requested sign) — rejected, looks like a parser error
- omit the event entirely — rejected, breaks audit honesty (a reduction was attempted; it just couldn't apply)
- separate event-type field — over-engineered for a string-based audit log

`(floored)` suffix on the existing string format keeps the audit log honest, parseable with simple `.Contains("(floored)")`, and doesn't change the event count.

`OpponentShadowTracker` does not exist as a separate class in `pinder-core` — only `SessionShadowTracker` is used (verified via repo-wide grep). No second clamp site needed.

## DoD Evidence

**Test tail** (`dotnet test --filter "Shadow"` — 392 tests in scope):
```
[xUnit.net 00:00:02.03]     Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Dread_T1 [SKIP]
[xUnit.net 00:00:02.03]     Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Madness_T2 [SKIP]
[xUnit.net 00:00:02.03]     Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Dread_T2 [SKIP]
[xUnit.net 00:00:02.03]     Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Denial_T2 [SKIP]
  Skipped Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Dread_T1 [1 ms]
  Skipped Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Madness_T2 [1 ms]
  Skipped Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Dread_T2 [1 ms]
  Skipped Pinder.Core.Tests.RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_Denial_T2 [1 ms]

Passed!  - Failed:     0, Passed:   380, Skipped:    12, Total:   392, Duration: 549 ms - Pinder.Core.Tests.dll (net8.0)
```

Full suite (`dotnet test` — 2508 tests): `Failed: 6, Passed: 2484, Skipped: 18`. The 6 failures are pre-existing environmental skips in `CharacterLoaderSpecTests` (`Assert.Fail("SKIPPED: Prompt directory 'design/examples' not found in repo")`) — they require disk fixtures that aren't checked into the submodule. Unrelated to this change; verified by inspection of the assertion text.

**Build tail** (`dotnet build src/Pinder.Core/Pinder.Core.csproj -c Release`):
```
    23 Warning(s)
    0 Error(s)

Time Elapsed 00:00:04.13
```
(All 23 warnings are pre-existing nullable-reference warnings in `GameSession.cs` — none introduced by this PR.)

**git log**:
```
a959815 fix(#405): floor shadow stats at 0 (read + write + restore clamps)
```

**Files changed**: 6 (1 source, 5 tests; +297 / -25)
- `src/Pinder.Core/Stats/SessionShadowTracker.cs` — clamps at three layers
- `tests/Pinder.Core.Tests/Issue405_ShadowFloorTests.cs` — 5 new regression tests (4 ACs + 1 read-time-clamp guard)
- `tests/Pinder.Core.Tests/ShadowGrowthEventTests.cs` — updated 1 stale test, added 1 counterpart for the real-reduction path
- `tests/Pinder.Core.Tests/ShadowGrowthSpecTests.cs` — updated 1 stale test (`Edge_NegativeShadowDelta_Allowed` → `_FlooredAtZero`)
- `tests/Pinder.Core.Tests/ShadowReductionSpecTests.cs` — updated 2 stale tests (`AC1_DateSecured_DreadCanGoNegative` → `_DreadAtZero_ReductionFlooredNotNegative`; `AC1_DateSecured_GrowthEventContainsDreadDateSecured` now pre-grows Dread so the reduction is real)
- `tests/Pinder.Core.Tests/ShadowReductionTests.cs` — updated 1 stale test (`DateSecured_DreadReductionCanGoNegative` → `_FlooredAtZero`)

The 4 stale tests previously asserted `GetDelta` could be `-1` etc. — i.e. they encoded the silent-state-corruption bug as expected behaviour. Each is rewritten to assert the new floor invariant and (where applicable) the `(floored)` audit-log marker.

**Acceptance criteria mapping** (from #405):
- [x] `GetEffectiveShadow(s) >= 0` for any sequence of `ApplyGrowth` / `ApplyOffset` — covered by `T1_ReductionsAtZeroFloor_EffectiveStaysAtZero` + `T2_NetReductionBelowZero_FloorsAtZero_WithFlooredAuditEvents`.
- [x] `GetEffectiveStat` not increased above base by negative shadow accumulation — covered by `T3_FlooredShadow_DoesNotBuffPairedStat` (plus `T3b` defends against a hypothetical bypass of the write clamp).
- [x] Regression test: `madness: 0`, three `-1` calls, effective stays 0 — `T1`.
- [x] Regression test: `madness: 1`, four `-1` calls, effective 0, audit log honest — `T2`.
- [x] Snapshot replay clamps too — `T4_RestoreFromSnapshot_NegativeTargetValue_ClampedToZero`.
- [x] Snapshot schema discipline: no new fields added to `GameSession` or `TurnSnapshot` — the audit-log marker is a string-format change inside an existing `growth events` list. `SessionSnapshot.cs` does not need updating.

## Research Log

| Topic | Source | Key finding |
|-------|--------|-------------|
| C# integer division of negatives | Existing `GetEffectiveStat` line `int penalty = totalShadow / 3;` and the issue body | C# `int` division truncates toward zero, so `floor(-2/3)` is `0` in C#, **not** `-1` as math.floor would produce — but the issue body was correct in the practical effect: `-2 / 3 == 0` in C#, but the problem is that `baseVal - 0 = baseVal` is still wrong (no penalty applied; threshold checks broken). For `shadow = -3`, `-3 / 3 == -1` and `baseVal - (-1) = baseVal + 1` — silent buff confirmed. The fix (clamp at 0) is correct regardless. |
| `RestoreFromSnapshot` semantics | `src/Pinder.Core/Stats/SessionShadowTracker.cs` + `GameSession.cs:340` | Used by `GameSession.RestoreState` for resimulation. Takes `Dictionary<string,int>` of effective values. Clamping the target value before computing the delta is the cheapest fix that preserves all callers. |
| `OpponentShadowTracker` existence | repo-wide `grep -rn "OpponentShadowTracker"` | Class does not exist; only `SessionShadowTracker` is referenced. The opponent uses the player-side type. No second clamp site needed. |
| Existing audit-log format | `ApplyOffset` description string and `ShadowReductionSpecTests` assertions | Existing format is `"{Shadow} {sign} ({reason})"`. Adding a `(floored)` suffix preserves backward-compatible parsing for any consumer that just substring-matches on the shadow name + sign. |

Closes #405

### DIFF
--- a/src/Pinder.Core/Stats/SessionShadowTracker.cs
+++ b/src/Pinder.Core/Stats/SessionShadowTracker.cs
@@ -7,6 +7,12 @@ namespace Pinder.Core.Stats
     /// Mutable shadow tracking layer wrapping an immutable StatBlock.
     /// Tracks in-session shadow growth deltas and provides effective stat values
     /// that account for session-accumulated shadow growth.
+    ///
+    /// Invariant (issue #405): effective shadow values are floored at 0.
+    /// Negative shadow values are not legal game state — they would silently
+    /// buff the paired positive stat via integer-division flooring in
+    /// <see cref="GetEffectiveStat"/>. The floor is enforced at three layers
+    /// (read, write, restore) for defense in depth.
     /// </summary>
     public sealed class SessionShadowTracker
     {
@@ -27,11 +33,13 @@ namespace Pinder.Core.Stats
         }
 
         /// <summary>
-        /// Returns the effective shadow value: base shadow + in-session delta.
+        /// Returns the effective shadow value: base shadow + in-session delta, floored at 0.
+        /// (issue #405: negative effective shadows are not legal — they would silently buff
+        /// the paired positive stat via floor-division in <see cref="GetEffectiveStat"/>.)
         /// </summary>
         public int GetEffectiveShadow(ShadowStatType shadow)
         {
-            return _baseStats.GetShadow(shadow) + GetDelta(shadow);
+            return Math.Max(0, _baseStats.GetShadow(shadow) + GetDelta(shadow));
         }
 
         /// <summary>
@@ -59,19 +67,24 @@ namespace Pinder.Core.Stats
 
         /// <summary>
         /// Returns the effective stat modifier accounting for in-session shadow growth.
-        /// Formula: baseStats.GetBase(stat) - floor((baseStats.GetShadow(pairedShadow) + delta[pairedShadow]) / 3)
+        /// Formula: baseStats.GetBase(stat) - floor(max(0, baseStats.GetShadow(pairedShadow) + delta[pairedShadow]) / 3)
         /// Uses StatBlock.ShadowPairs to determine the paired shadow stat.
+        /// (issue #405: shadow component is floored at 0 to prevent negative shadows from
+        /// silently buffing the paired stat via integer-division of negatives.)
         /// </summary>
         public int GetEffectiveStat(StatType stat)
         {
             var pairedShadow = StatBlock.ShadowPairs[stat];
-            int totalShadow = _baseStats.GetShadow(pairedShadow) + GetDelta(pairedShadow);
+            int totalShadow = Math.Max(0, _baseStats.GetShadow(pairedShadow) + GetDelta(pairedShadow));
             int penalty = totalShadow / 3;
             return _baseStats.GetBase(stat) - penalty;
         }
 
         /// <summary>
         /// Returns only the in-session delta for a shadow stat (0 if no growth has occurred).
+        /// Note: this is the RAW stored delta — it is bounded below such that
+        /// base + delta &gt;= 0, but does not by itself indicate a clamp event. Use
+        /// <see cref="GetEffectiveShadow"/> for the user-visible value.
         /// </summary>
         public int GetDelta(ShadowStatType shadow)
         {
@@ -82,20 +95,47 @@ namespace Pinder.Core.Stats
         /// <summary>
         /// Applies a signed delta to a shadow stat. Unlike ApplyGrowth, allows negative values
         /// (e.g., Fixation −1 offset for using 4+ different stats). Stores a description event.
+        ///
+        /// (issue #405) The effective shadow value is floored at 0. If the requested negative
+        /// delta would drive the effective value below 0, only the portion that brings the
+        /// shadow down to 0 is applied; the remainder is recorded as a "(floored)" event so
+        /// the audit log honestly reflects what actually happened.
         /// </summary>
         /// <param name="shadow">The shadow stat to adjust.</param>
         /// <param name="delta">Signed delta (positive = growth, negative = reduction).</param>
         /// <param name="reason">Human-readable reason for the change.</param>
-        /// <returns>Description string: "{ShadowStatName} {+/-delta} ({reason})"</returns>
+        /// <returns>Description string: "{ShadowStatName} {+/-applied} ({reason})" — or
+        /// "{ShadowStatName} +0 ({reason}) (floored)" when the reduction was fully suppressed.</returns>
         public string ApplyOffset(ShadowStatType shadow, int delta, string reason)
         {
-            if (_deltas.ContainsKey(shadow))
-                _deltas[shadow] += delta;
-            else
-                _deltas[shadow] = delta;
+            int currentEffective = _baseStats.GetShadow(shadow) + GetDelta(shadow);
+            // Pre-clamp: if the running delta has been suppressed in the past, treat the
+            // floor as 0 for the purposes of this call. (currentEffective is already
+            // monotone with stored delta; we just guard against drifting below 0 from here.)
+            if (currentEffective < 0) currentEffective = 0;
 
-            string sign = delta >= 0 ? $"+{delta}" : delta.ToString();
-            string description = $"{shadow} {sign} ({reason})";
+            int appliedDelta = delta;
+            bool floored = false;
+
+            if (delta < 0 && currentEffective + delta < 0)
+            {
+                // Only apply enough to take effective down to 0; record the rest as floored.
+                appliedDelta = -currentEffective; // 0 or negative; brings effective to exactly 0
+                floored = true;
+            }
+
+            if (appliedDelta != 0)
+            {
+                if (_deltas.ContainsKey(shadow))
+                    _deltas[shadow] += appliedDelta;
+                else
+                    _deltas[shadow] = appliedDelta;
+            }
+
+            string sign = appliedDelta >= 0 ? $"+{appliedDelta}" : appliedDelta.ToString();
+            string description = floored
+                ? $"{shadow} {sign} ({reason}) (floored)"
+                : $"{shadow} {sign} ({reason})";
             _growthEvents.Add(description);
             return description;
         }
@@ -105,6 +145,10 @@ namespace Pinder.Core.Stats
         /// For each shadow stat, sets delta = targetValue − baseStats.GetShadow(shadow).
         /// Clears any previously accumulated deltas and the growth event log before restoring.
         /// Used by GameSession.RestoreState when resimulating from a snapshot.
+        ///
+        /// (issue #405) Restored target values are clamped at 0. Old or corrupt snapshots
+        /// containing negative effective values are normalised to 0 on restore so the floor
+        /// invariant holds across resimulation.
         /// </summary>
         /// <param name="targetValues">Map of ShadowStatType.ToString() → desired effective value.</param>
         public void RestoreFromSnapshot(Dictionary<string, int> targetValues)
@@ -118,8 +162,9 @@ namespace Pinder.Core.Stats
                 string key = shadow.ToString();
                 if (targetValues.TryGetValue(key, out int targetValue))
                 {
+                    int clampedTarget = Math.Max(0, targetValue);
                     int baseValue = _baseStats.GetShadow(shadow);
-                    int delta = targetValue - baseValue;
+                    int delta = clampedTarget - baseValue;
                     if (delta != 0)
                         _deltas[shadow] = delta;
                 }
--- a/tests/Pinder.Core.Tests/Issue405_ShadowFloorTests.cs
+++ b/tests/Pinder.Core.Tests/Issue405_ShadowFloorTests.cs
@@ -1,0 +1,175 @@
+using System.Collections.Generic;
+using Pinder.Core.Stats;
+using Xunit;
+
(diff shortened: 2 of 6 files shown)


decay256 (commented):
## Code Review

**Maturity level:** mvp (production engine module)
**Blocking issues:** 0
**Non-blocking issues:** 1

**Verdict: APPROVE**

(Posting as `--comment` per LESSONS_LEARNED §12 — gh token is the PR author identity, formal `--approve` is blocked.)

### Verified against acceptance criteria from issue #405

- ✅ `GetEffectiveShadow(s) >= 0` for any sequence of `ApplyGrowth`/`ApplyOffset`/`RestoreFromSnapshot` (read clamp at line 41, write clamp in `ApplyOffset`, restore clamp in `RestoreFromSnapshot`).
- ✅ `GetEffectiveStat(stat)` cannot be increased above its base by negative shadow accumulation. Clamp lands BEFORE the `/3` integer division (line 75-77 — correct order).
- ✅ Reductions on a 0 effective shadow are written as `appliedDelta = 0` and stored as a no-op (no `_deltas` mutation when `appliedDelta == 0`), with the audit log carrying `(floored)`.
- ✅ Audit-log convention: `(floored)` suffix is applied **only** when `floored == true`, which is set only in the partial-or-full-suppression branch. Verified consistent across the new and updated tests (`T2_NetReductionBelowZero…`, `ApplyOffset_AddsEvent`, `ApplyOffset_NegativeWithinFloor_AddsRealReductionEvent`, `AC1_DateSecured_DreadAtZero_ReductionFlooredNotNegative`, `AC1_DateSecured_GrowthEventContainsDreadDateSecured` which now negative-asserts `!Contains("(floored)")`).
- ✅ `RestoreFromSnapshot` clamps target via `Math.Max(0, targetValue)` (line 165) — old/corrupt snapshots normalise on restore.

### Verified hazards specific to this PR

1. **Three-layer clamp consistency.** Read clamp (`GetEffectiveShadow`), read clamp (`GetEffectiveStat`), write clamp (`ApplyOffset`), restore clamp (`RestoreFromSnapshot`). The stored `_deltas[shadow]` is the **post-clamp applied delta** (not the requested delta) — so `GetDelta` returns the truthful applied amount. The read-time clamps are pure defense-in-depth and cannot disagree with the stored value because the stored value is already in-bounds. The XML comment on `GetDelta` explicitly says "this is the RAW stored delta — bounded below such that base + delta >= 0" — that is exactly the contract the code now enforces.
2. **`ApplyGrowth` parity.** `ApplyGrowth` throws `ArgumentOutOfRangeException` on `amount <= 0` (line 56-57). Cannot be a vector for negative state. ✓
3. **Audit-log honesty.** When `delta=-1` hits a floor (currentEffective=0), `appliedDelta=0`, description becomes `"{Shadow} +0 ({reason}) (floored)"`. Test `T2` asserts this exact behavior across 3 successive floored events plus 1 real one. The pre-existing `ApplyOffset_AddsEvent` was correctly updated from "asserts `-1` in description" (false post-fix because base=0 → floored) to "asserts `(floored)` marker", and a new `ApplyOffset_NegativeWithinFloor_AddsRealReductionEvent` covers the not-floored real-reduction path that the original test was clearly intending to verify.
4. **Stale tests updated correctly.** All 4 pre-existing tests that encoded the bug (`ShadowGrowthEventTests.ApplyOffset_AddsEvent`, `ShadowGrowthSpecTests.Edge_NegativeShadowDelta_FlooredAtZero` (renamed from `Edge_NegativeShadowDelta_Allowed`), `ShadowReductionSpecTests.AC1_DateSecured_DreadAtZero_ReductionFlooredNotNegative` (renamed from `AC1_DateSecured_DreadCanGoNegative`), `ShadowReductionTests.DateSecured_DreadReduction_FlooredAtZero` (renamed from `DateSecured_DreadReductionCanGoNegative`)) were inspected against `origin/main`. In every case the change is "expected value updated to match the corrected invariant" — none are deletions, none are inverted-meaning edits, none are weakened to silently-pass. The renames also surface the contract change in the test catalog.
5. **`AC1_DateSecured_GrowthEventContainsDreadDateSecured`** correctly added an `ApplyGrowth(Dread, 2, "setup")` pre-stage so the `-1` reduction is real (not floored), preserving the original test's intent of verifying the real-reduction event format. Adjacent test `AC1_DateSecured_DreadAtZero_ReductionFlooredNotNegative` covers the floored-from-zero case.
6. **`OpponentShadowTracker` non-existence verified.** `grep -rn "OpponentShadowTracker\|OpponentShadows\|_opponentShadows"` confirms `_opponentShadows` is `SessionShadowTracker?` (same class) in `GameSession.cs:37` and `GameSessionConfig.cs:20`. The fix applies to opponent shadows transitively via the same code path. No separate type exists.
7. **Order of operations in `GetEffectiveStat`.** Verified line-by-line: `int totalShadow = Math.Max(0, baseShadow + delta);` then `int penalty = totalShadow / 3;`. Clamp lands BEFORE the divide — the canonical hazard from the issue (negative integer division flooring "downward" producing buffs) cannot recur.

### Cross-cutting

- ✅ Schema discipline: `git diff origin/main..HEAD --name-only` shows only `src/Pinder.Core/Stats/SessionShadowTracker.cs` + 5 test files. No `TurnResult` / `GameSession` / `TurnSnapshot` mutation. No new fields. No `SessionSnapshot.cs` comment-block change required.
- ✅ Logging: format-string change to existing `_growthEvents` description only. No new log lines, no new sinks.

### Tests soundness

- 5 new tests in `Issue405_ShadowFloorTests.cs` (T1, T2, T3, T3b, T4) cover: (1) reductions on 0-shadow, (2) net reduction below 0 + audit honesty, (3) negative shadow does not buff paired stat (the canonical bug from the issue body), (3b) read-time clamp guard against hypothetical bypass, (4) restore-from-snapshot clamps. The full set matches the prompt checklist.
- The 4 updated stale tests are real corrections, not edits-to-pass.

### Build + tests

```
$ dotnet build src/Pinder.Core/Pinder.Core.csproj -c Release
    23 Warning(s)
    0 Error(s)
Time Elapsed 00:00:03.53
```
(All 23 warnings are pre-existing nullable-ref warnings in `GameSession.cs`, unrelated to this PR.)

```
$ dotnet test tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj -c Release --filter "Shadow"
Passed!  - Failed: 0, Passed: 380, Skipped: 12, Total: 392, Duration: 1 s
```
(12 skips are pre-existing `RulesSpec.RulesSpecTests.Rule_S7_ShadowTaint_*` skips, unrelated.)

### Non-blocking

- **`tests/Pinder.Core.Tests/ShadowGrowthSpecTests.cs:1078-1080`** — the second assertion in `Edge_NegativeShadowDelta_FlooredAtZero`:
  ```csharp
  Assert.True(tracker.GetEffectiveShadow(ShadowStatType.Fixation) == 0
      && tracker.GetDelta(ShadowStatType.Fixation) >= -tracker.GetEffectiveShadow(ShadowStatType.Fixation) - 0,
      "#405: stored delta must not drive base+delta below 0");
  ```
  After the prior `Assert.Equal(0, GetEffectiveShadow(...))`, this expression simplifies to `0 == 0 && GetDelta >= 0`, i.e. effectively `GetDelta >= 0` — and given that `MakeTracker()` has base shadow 0 and the prior `ApplyGrowth(2)` then `ApplyOffset(-3)` flooring leaves stored delta at 0, this is a tautology piggy-backing on the previous assertion. Suggest replacing with a single direct assertion (e.g. `Assert.Equal(0, tracker.GetDelta(ShadowStatType.Fixation));` or `Assert.True(tracker.GetDelta(ShadowStatType.Fixation) >= 0);`). Cosmetic only — does not affect correctness or coverage.

### Positive notes

- The decision to **store the post-clamp applied delta** (rather than the requested delta with a separate "floored" flag) is the right shape: `GetDelta` returns truthful applied state, and the audit log records what actually happened. The redundant read-time clamp is honest defense-in-depth, not paper over a leaky write path.
- The `(floored)` suffix is a clean string-format choice — easy to grep, doesn't break parsing of existing prefixes, and is asserted positively in the new tests.
- All 4 stale tests were *renamed* to surface the contract change (e.g. `…CanGoNegative` → `…FlooredAtZero`, `…CanGoNegative` → `…FlooredNotNegative`, `Allowed` → `FlooredAtZero`). Future readers will not re-introduce the bug under the old test name.
- XML doc on `GetDelta` explicitly warns "this is the RAW stored delta… does not by itself indicate a clamp event. Use GetEffectiveShadow for the user-visible value" — exactly the kind of contract pin that prevents a future caller from short-circuiting through the wrong API.

**Verdict: APPROVE**

PR #393, opened 2026-04-03
[Issue #350] Spec: Session runner shadow tracking via GameSessionConfig

decay256:
Refs #350

## Summary
Specification document for enabling shadow tracking in the session runner by passing `GameSessionConfig` with `PlayerShadows`.

Covers:
- Wiring `SessionShadowTracker(sableStats)` into `GameSession` via `GameSessionConfig`
- Per-turn shadow growth event display (`⚠️ SHADOW GROWTH: ...`)
- Session-end shadow delta summary table
- All 5 acceptance criteria specified with edge cases and error conditions

## Spec Location
`docs/specs/issue-350-spec.md`

## DoD Evidence
**Branch:** issue-350-spec-shadow-tracking
**Commit:** a75a310

Issue #16, opened 2026-03-22
Vision: Issue #7 dependency on #4 is unresolvable — #4 is a vision-concern, not a code PR


decay256:
## Concern
Issue #7 lists `Depends on: #1, #2, #4`. Issues #1 and #2 are merged PRs — correct. But #4 is a vision-concern issue ("Issue #1 will break test suite"), not a code change. It has no branch, no PR, and will never be "merged." The dependency is unresolvable as stated.

Additionally, the **real** untracked dependency is the success scale implementation. Issues #8 and #15 both flag this: #7 specifies tests for §5 success scale (Interest +1/+2/+3/+4 based on margin), but no code exists to produce these values. `RollResult` has no `SuccessMargin` property. `RollEngine` treats all successes as `FailureTier.None` with no differentiation.

## Why it matters
- An automated sprint system checking "is #4 merged?" will never unblock #7.
- Even if manually unblocked, the agent implementing #7 cannot write success scale assertions — there's nothing to assert against.
- This is the second sprint where this gap has been flagged (#8, #15) and not resolved.

## What needs to happen
1. **Fix #7's dependency list**: Replace `#4` with the actual prerequisite (a success scale implementation issue, or remove if none needed).
2. **Resolve the success scale gap** before #7 can be implemented. Either:
   - (a) Add a new issue to this sprint: implement `SuccessScale` (SuccessMargin on RollResult, interest delta calculation). Make #7 depend on it.
   - (b) Remove §5 success scale assertions from #7's acceptance criteria and track them in a future sprint.
3. See also: #8, #15 (prior flags on the same gap).

**Role**: backend-engineer
**Maturity**: prototype

decay256:
Closing: stale vision concern from early sprint coordination. Architecture has evolved significantly since this was filed.

PR #798, opened 2026-05-02
refactor(#790): GameSession.Clone() — engine forking for fast-gameplay (Phase 4)

### LINKED ISSUE #790: [refactor] Phase 4 — GameSession.Clone() (engine forking)
> Phase 4 of the fast-gameplay epic (#393). Plan PR #422 §7.

## Goal

Implement `GameSession.Clone()` (or equivalent via snapshot-roundtrip) so the fast-gameplay scheduler in Phase 5 can fork three independent game sessions from one parent before the player click.

## Why

This is the engineering core of fast-gameplay: three branches need independent mutable state. Once Phases 1–3 land:
- `_opponentHistory` is in `GameSession` (Phase 1) → cloneable
- RNGs are pre-rolled (Phase 2) → no shared mutable RNG state to clone
- Audit sinks are per-branch (Phase 3) → not on `GameSession`

So clone is mostly a matter of deep-copying the Category-A fields enumerated in plan v2 §2.

### Path 1 — explicit field-by-field `Clone()`
Add a `public GameSession Clone()` method that constructs a new `GameSession` and deep-copies every Category-A field.

Pros: fast, allocations are minimized.
Cons: every new field on `GameSession` must remember to extend `Clone()`. Easy to forget.

### Path 2 — snapshot-and-restore
Use the existing `CreateSnapshot()` + `RestoreState()` machinery: take a snapshot, instantiate a fresh `GameSession` with the same construction params, restore the snapshot. The replay path already does this; reusing it gets correctness for free.

Pros: already tested; "what if a new field is missing from clone" is impossible because the snapshot path catches it (replay tests fail).
Cons: serialize-deserialize round-trip on every clone (2 ms? 10 ms? measure). Negligible vs LLM latency but not free.

**Plan v2 recommends starting with Path 2**. If profiling shows the round-trip adds material latency to the option-display path, fall back to Path 1.

## Acceptance criteria

- [ ] `GameSession.Clone()` (or equivalent) returns a fully-independent session that mutates without affecting the original.
- [ ] Property test: `GameSession g; var c = g.Clone(); g.RunSomeMutation(); c.State == g.PreMutationState;` (clone independence).
- [ ] Property test: running the same option through the original and through the clone produces byte-identical post-state (assuming same `PerOptionDicePool`).
- [ ] All Phase-0 regression tests still pass.
- [ ] Path 1 vs Path 2 decision documented in the PR body, with profiling numbers if Path 2 chosen.
- [ ] If new fields were added to `GameSession` since plan v2 was written, they're handled correctly.
- [ ] After `#786` (energy removal) lands, this phase has no clock-clone story to write.

## Estimated effort

T-shirt: **L**. ~+220 LOC. Heart of the fast-gameplay engineering.

## Hard prerequisites

- #786 (energy removal) — simplifies clock-clone story
- Phase 0 (#787 pinder-core)
- Phase 1 (#788 pinder-core)
- Phase 2 (#789 pinder-core)
- Phase 3 (#424 pinder-web)

## Related

- Epic: #393
- Plan: PR #422 §7

decay256:
Wave 4 of the #393 fast-gameplay drain. Adds public `GameSession.Clone()` returning a fully-independent session for the fast-gameplay scheduler (#425, Phase 5) to fork three branches per turn before the player commits.

Closes #790.

## Path 1 vs Path 2 — decision documented

Plan v2 §7 recommended **Path 2** (snapshot-and-restore round-trip) because "replay tests would catch missing fields." Audit of the existing `CreateSnapshot` / `RestoreState` machinery showed it does **not** cover all of `GameSession`'s mutable state:

| Field | In `RestoreState`? |
|---|---|
| `InterestMeter` | ✅ via `TargetInterest` |
| `_momentumStreak`, `_turnNumber` | ✅ |
| `TrapState` | ✅ |
| `_history` | ✅ |
| `_opponentHistory` | ✅ |
| `_comboTracker` | ✅ |
| `_rizzCumulativeFailureCount` | ✅ |
| `_playerShadows` (effective values) | ✅ |
| `_xpLedger` | ❌ |
| `_horninessRoll`, `_horninessTimeModifier`, `_sessionHorniness` | ❌ |
| `_opponentShadows` | ❌ |
| `_shadowGrowthEvaluator` (counters: `_honestySuccessCount`, `_saUsageCount`, `_charmUsageCount`, `_charmMadnessTriggered`, `_saOverthinkingTriggered`, `_statsUsedPerTurn`, `_highestPctOptionPicked`) | ❌ |
| `_topics` | ❌ |
| `_lastStatUsed`, `_activeWeakness`, `_activeTell` | ❌ |
| `_pendingCritAdvantage`, `_pendingMomentumBonus` | ❌ |
| `_shadowDisadvantagedStats`, `_currentShadowThresholds` | ❌ |
| Mid-turn `_currentOptions`, `_currentDicePools`, `_currentHasAdvantage`, `_currentHasDisadvantage`, `_injectedNextPool` | ❌ |
| `_ended`, `_outcome` | ❌ (`MarkEnded` is separate) |

Path 2 would have required expanding both `CreateSnapshot` and `RestoreState` to full coverage — a much larger blast radius on the production replay path. **Path 1** (explicit private clone constructor) chosen for narrower scope. The reflection-based completeness test guards against the Path 1 risk ("forgot to extend Clone in a future PR").

No profiling numbers needed — Path 2 was rejected on architectural grounds (incomplete scope), not perf.

## Implementation

- **New private `GameSession(GameSession src)` clone constructor**, with three field categories explicitly documented in XML:
  - **Shared by reference** (immutable / stateless / pure adapters): `_player`, `_opponent`, `_llm`, `_dice`, `_trapRegistry`, `_clock`, `_rules`, `_statDeliveryInstructions`, `_onTextLayerNoop`, `_globalDcBias`.
  - **Deep-copied** (mutable engine state): `InterestMeter`, `TrapState`, `ComboTracker`, `XpLedger`, `SessionShadowTracker × 2`, `ShadowGrowthEvaluator`, `_topics`, `_history`, `_opponentHistory`, `_shadowDisadvantagedStats`, `_currentShadowThresholds`, `_currentOptions`, `_currentDicePools`.
  - **Deep-cloned RNG state**: `System.Random` instances cloned via reflection-based seed-array copy in `RandomCloner`. The steering RNG is shared between `SteeringEngine` and `HorninessEngine` (per the public ctor's wiring); the clone preserves that sharing shape. `_statDrawRng` is independent.
- Added `Clone()` methods to `InterestMeter`, `TrapState`, `ComboTracker`, `XpLedger`, `SessionShadowTracker`, `ShadowGrowthEvaluator`. Each carries the same "mutating either side does not affect the other" contract in XML.
- Public `GameSession.Clone()` exposes the constructor with a full XML doc covering use case, Path 1/2 decision, and the three locked properties.

### `RandomCloner`

`System.Random`'s `_impl` wraps a `CompatPrng` struct that holds a reference to a backing `_seedArray`. Naive `MemberwiseClone` shares the array silently — clones drift after the first `.Next()` on the parent. The helper:

1. `MemberwiseClone`s the impl.
2. For every value-type field on the impl, boxes it and replaces nested array fields with `.Clone()` copies. (Specifically: walks `CompatPrng._seedArray` and clones it.)
3. `MemberwiseClone`s the outer `Random` and grafts the cloned impl.

`RandomClonerTests` pins three properties: identical next-N at clone-time, parent advancement does not shift clone, clone advancement does not shift parent.

## Tests (Category=Phase4, 10 tests, all green)

| Test | Property |
|---|---|
| `Clone_IsIndependent_FromParentMutationsOnConversationHistory` | Independence: parent runs a turn → clone's snapshot, history, opponent history all unchanged |
| `Clone_IsIndependent_OnTopicsList` | Independence: `AddTopic` on parent does not bleed into clone |
| `Clone_IsDeterministic_OnInjectedDicePool` | Determinism: same option through parent + clone with same `PerOptionDicePool` produces byte-identical post-state |
| `Clone_CoversEveryGameSessionField` | **Reflection-based completeness** — every instance field is either on the documented `sharedByReference` allowlist OR is value-equal OR is a different instance |
| `Clone_SnapshotMatchesParent_AtCloneTime` | Unmutated clone produces byte-equal `CreateSnapshot` output |
| `Clone_HorninessSessionRoll_PreservedExactly` | The ctor's d10 horniness roll is copied verbatim (Path 2 round-trip would have re-drawn) |
| `RandomCloner` × 4 | Identical next-sequence; independence in both directions; null-arg throws |

### Mutation-tests verifying the completeness assertion is real (not a no-op)

Per LESSONS_LEARNED #16 (negative-grep tests must be seen failing once before they're trustworthy), I mutated the clone constructor twice during development:

1. **Forgot to deep-copy `_topics`** (`_topics = src._topics;` instead of `new List<>(src._topics)`):
   ```
   Failed: _topics: shared by reference after clone, but is not on the
   documented allowlist. Either deep-copy this field in GameSession.Clone,
   or add it to the sharedByReference set in this test with a one-line
   justification in the Clone constructor's XML doc.
   ```
2. **Forgot to copy a value-type field** (`_horninessRoll = 0;`):
   ```
   Failed: _horninessRoll: value-type field differs after clone (parent=5, clone=0)
   ```

Both mutations were caught by `Clone_CoversEveryGameSessionField` with actionable diagnostics. Reverted before commit.

## Cross-repo

Pinder-core only. The pinder-web submodule pointer bump will follow in a separate parent-repo PR. **`BranchScheduler` (#424) is intentionally not modified** — that's #425's scope (Phase 5 will consume `Clone()`).

### API constraints for the scheduler

- `Clone()` works at any point in the session lifecycle: pre-`StartTurnAsync`, post-`StartTurnAsync` / pre-`ResolveTurnAsync` (mid-turn carry-over is part of the clone surface), and post-`ResolveTurnAsync`. The fast-gameplay scheduler will use the post-`StartTurnAsync` clone shape: parent runs `StartTurnAsync` once, then forks three clones, each gets its own `InjectNextDicePool` for a different option, then all three run `ResolveTurnAsync` in parallel.
- The cloned `_dice` is shared by reference with the parent. The scheduler MUST NOT call `_dice` on cloned sessions — instead it MUST call `InjectNextDicePool(pool)` per-branch before `ResolveTurnAsync`. The `PerOptionDicePool` is immutable and safe to share across branches if they happen to be wired with the same option.
- The cloned LLM adapter is shared by reference. Production scheduling expects each branch to have its own adapter chain (separate transport + per-branch audit sink, per Phase 3). When wiring, replace `_llm`-equivalent host-side state per-branch — the engine's `Clone()` doesn't know about audit sinks (that's pinder-web's concern).
- The cloned RNG state is fork-independent. Three parallel branches can call `_steeringEngine` / `_horninessEngine` / `_statDrawRng` without racing each other.

## Verification

- **Phase 0**: 31/31 passing (no regression)
- **Full `Pinder.Core.Tests`**: 2508 passing (was 2498, +10), 6 pre-existing `CharacterLoaderSpec` flakes unchanged, 18 skipped
- **Release build**: clean


decay256 (commented):
## Code Review (no-context)

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 1
**Verdict**: **CLEAR_TO_MERGE**

### Verification matrix

| Check | Result |
|---|---|
| Phase 0 regression suite (`Category=Phase0`) | **31 / 31 green** |
| Phase 4 clone suite (`Category=Phase4`) | **10 / 10 green** |
| RandomCloner unit tests | **4 / 4 green** |
| GameApi tests (cross-repo, against bumped pointer) | **536 / 536 green** (28 s) |
| `dotnet build Pinder.sln` (whole-tree, web ↔ core) | **0 errors, 0 warnings** |
| Reviewer-injected mutation on `Clone_CoversEveryGameSessionField` | **Caught with actionable diagnostic** ✅ |

### Mutation test (re-verified by reviewer)

To confirm the reflection-based completeness assertion is non-vacuous, I edited `GameSession.cs` line 275:

```csharp
- _topics = new List<CallbackOpportunity>(src._topics);
+ _topics = src._topics; // share by reference
```

`Clone_CoversEveryGameSessionField` failed with:

```
_topics: shared by reference after clone, but is not on the documented
allowlist. Either deep-copy this field in GameSession.Clone, or add it
to the sharedByReference set in this test with a one-line justification
in the Clone constructor's XML doc.
```

This is the right shape: the diagnostic tells the next maintainer exactly what to do. Reverted via `git checkout`.

### `RandomCloner` — the subtle bit

Confirmed:
- Throws `InvalidOperationException` when `_impl` field can't be located via reflection — refuses to silently fall back to a fresh-seeded `Random` (which would silently break clone determinism). ✅
- Walks fields generically by type (`IsArray`, `IsValueType` recursion) rather than hardcoded BCL field names — future .NET 8.x patch revisions won't silently re-couple seed arrays. ✅
- `RandomClonerTests.Clone_IsIndependent_FromParentAdvancement` exercises exactly the contract the doc-comment claims: parent advances 100 calls after clone, clone still produces the originally-predicted next-N. ✅
- `Clone_IsIndependent_FromCloneAdvancement` covers the symmetric direction (clone advances, parent untouched). ✅

### `GameSession.Clone()` — Phase 5 API constraints

Spot-checked against the `private GameSession(GameSession src)` ctor (lines 255–325):

- `_dice` is **shared by reference** (line 261) — Phase 5 scheduler must use `InjectNextDicePool` per branch, as documented. ✅
- `_llm` is **shared by reference** (line 260) — Phase 5 wires per-branch decoration via #424's BranchScheduler. ✅
- `_steeringRng` and `_statDrawRng` are **deep-cloned via `RandomCloner.Clone(...)`** (lines 287–291) — three concurrent clones cannot race on shared seed state. ✅
- The `_steeringRng`-shared-with-`_horninessEngine` quirk from the public ctor is preserved on the clone path (single cloned RNG passed to both engine ctors), so behavioural shape is identical. ✅

### Path 1 vs Path 2 — rationale audit

I read `GameSessionHelpers.CreateSnapshot` (line 92): it captures only `interest, state, momentumStreak, traps, turnNumber, tripleBonusActive, opponentHistory`. It does **not** capture: `_xpLedger`, `_horninessRoll`, `_horninessTimeModifier`, `_sessionHorniness`, `_topics`, `_lastStatUsed`, `_pendingMomentumBonus`, `_pendingCritAdvantage`, `_rizzCumulativeFailureCount`, `_playerShadows`, `_opponentShadows`, `_shadowGrowthEvaluator`, `_comboTracker`, `_steeringRng`, `_statDrawRng`, `_currentOptions`, `_currentDicePools`, `_injectedNextPool`.

Path 2 (snapshot-and-restore round-trip) would have required expanding both `CreateSnapshot` and `RestoreState` to ~15+ additional fields — a much larger blast radius on the production replay path. **Path 1 was the correct call**, and the reflection-based completeness pin neutralises the only real downside ("future PR adds field, forgets to extend Clone").

The dedicated `Clone_HorninessSessionRoll_PreservedExactly` test specifically pins one of the fields Path 2 would have lost — the d10 horniness session roll computed once in the public ctor — making the rationale concrete rather than abstract.

### Independence + determinism

- `Clone_IsIndependent_FromParentMutationsOnConversationHistory` and `Clone_IsIndependent_OnTopicsList` cover the mutate-parent / clone-unchanged direction. ✅
- `Clone_IsDeterministic_OnInjectedDicePool` (lines 121–214) covers the same-option / same-dice / byte-identical-snapshot direction. ✅
- `Clone_SnapshotMatchesParent_AtCloneTime` is a fast-failing sanity pin: snapshot equality before any mutation. Good defensive layering. ✅

### Non-blocking note

- `_injectedNextPool` is a value-type-ish reference-shared field (line 325 comment "PerOptionDicePool: see clone-safety note below"). It's shared by reference between parent and clone, but `PerOptionDicePool` appears to be effectively immutable in the current API. The doc-comment promise inside the clone ctor and the completeness allowlist are consistent. If `PerOptionDicePool` ever grows mutable state, the existing reflection test will *not* automatically catch it (because the field is value-type and value-type comparison succeeds when the inner reference is the same). Worth a follow-up note in `LESSONS_LEARNED.md` so future PerOptionDicePool changes trigger a clone audit. **Non-blocking — does not change current correctness.**

### Positive notes

- The `private GameSession(GameSession src)` ctor groups fields by category (shared / deep-copied / RNG / value-type / mid-turn carry-over) with section banners — readable at a glance, hard to miss a category when adding a new field.
- The XML doc on the public `Clone()` method spells out (a) the use case, (b) the three locked invariants, (c) the Path-1-vs-Path-2 decision and rationale. This is the right level of inline documentation for an architecturally-load-bearing method.
- `Clone_HorninessSessionRoll_PreservedExactly` is exactly the right shape of test: it pins a *specific concrete failure mode* of the rejected alternative (Path 2), not just the abstract contract.
- Engine subcomponents (`InterestMeter`, `TrapState`, `ComboTracker`, `XpLedger`, `SessionShadowTracker`, `ShadowGrowthEvaluator`, `SteeringEngine`) all grew their own `Clone()` methods in this PR rather than being deep-copied via `JsonSerializer` round-trip — this is the right separation: each component owns its own clone semantics.

**Verdict: CLEAR_TO_MERGE** (orchestrator: re-bump pinder-web#457 to the post-squash core SHA before merging the web pointer.)

Issue #819, opened 2026-05-08
[character-assets] Pinder.RemoteAssets — Eigencore wrapper impl of IRemoteCharacterStore (#264 sub: E)


decay256:
# [character-assets] Pinder.RemoteAssets — Eigencore wrapper impl of IRemoteCharacterStore (#264 sub: E)

Parent epic: #264. Depends on #817 and #818 (both merged). Spec doc now reflects the deployed eigencore wire (refresh PR #852 / ticket #851 merged 2026-05-13).

## Status

✅ **Unblocked.** Eigencore-side #240 shipped; spec refresh (#851) merged. Ready to pick up via swarm-drain on the three sub-PRs listed under "Planned PR split" below.

## Goal

Provide the first concrete implementation of `IRemoteCharacterStore` (defined in #817), backed by Eigencore's asset-catalog HTTP API (per the refreshed `docs/specs/character-asset-vocabulary.md`). This is the only place in the codebase that knows Eigencore exists.

## Architectural rule (non-negotiable)

**Eigencore is a third-party app from Pinder's perspective.** Pinder-core has zero `using Eigencore.*` references. `Pinder.RemoteAssets` is the only assembly that knows eigencore exists. Wire vocabulary changes go via `asset_kind` discriminator (`character/v1` → `character/v2`), not by mutating in place. When Pinder and eigencore disagree, default fix is Pinder-side; an eigencore-side ticket only happens if "another game built on eigencore would also want this change" — see `AGENTS.md` for the full rule.

## Wire contract (post-#851 refresh)

The implementer MUST read `docs/specs/character-asset-vocabulary.md` end-to-end before opening sub-PR 1. Highlights:

- **Route prefix `/api/v1/assets`.** Absorb into the wrapper's `BaseUrl` config (don't hardcode `/api/v1` into every method).
- **Wire identity is `asset_id` (UUID), NOT `character_id`.** The C# POCO field stays `CharacterId` for Pinder-side clarity; `Pinder.RemoteAssets` is the only place that does the boundary rename.
- **Query param `tag` (singular, repeatable).** `?tag=a&tag=b`. **Not** `?tags=`. Eigencore (FastAPI) silently drops unknown query params — getting this wrong returns unfiltered results, not a 422.
[...]

Issue #817, opened 2026-05-08
[character-assets] IRemoteCharacterStore interface + query/filter contract (no impl) (#264 sub: D)


decay256:
# [character-assets] IRemoteCharacterStore interface + query/filter contract (no impl) (#264 sub: D)

Parent epic: #264. Depends on **#816**.

## Goal

Define `IRemoteCharacterStore`: the contract Unity (and future pinder-web "browse community characters" features) program against to fetch characters from a server. **Interface and types only — no implementation.** First implementation is #E, which is gated on Eigencore-side work and lives in a follow-up sprint.

## Context

Once #C lands, `ICharacterStore` covers local-disk and any in-memory test fakes. Remote stores have one extra concern: **discovery / querying** — Unity doesn't know the `character_id`s a server has; it needs to ask "show me public characters tagged 'official-pack'."

`IRemoteCharacterStore` extends `ICharacterStore` with a query surface. By keeping the load/save/delete/exists shape identical to `ICharacterStore`, Unity code that already takes an `ICharacterStore` works unchanged when handed a remote store.

## Requirements

Functional:
- `IRemoteCharacterStore : ICharacterStore` interface in `Pinder.Core.Characters`.
- New types in `Pinder.Core.Characters`:
  - `CharacterAssetMetadata` — opaque-to-engine metadata accompanying a remote character. Properties: `CharacterId`, `OwnerId`, `Tags` (`IReadOnlyList<string>`), `IsPublic`, `CreatedAt`, `UpdatedAt`, `AssetKind` (always `"character/v1"` for now). **The full vocabulary is specified in #F.**
  - `CharacterAssetQuery` — filter struct. Properties: `OwnerId` (optional), `Tags` (optional, all-must-match), `IsPublic` (optional), `Limit` (default 50, max 500), `Cursor` (optional, opaque pagination token).
  - `CharacterAssetPage` — return type. Properties: `Items` (list of `CharacterAssetMetadata`), `NextCursor` (nullable string).
- New methods on `IRemoteCharacterStore`:
  - `Task<CharacterAssetPage> QueryAsync(CharacterAssetQuery query, CancellationToken ct = default)` — returns metadata only, NOT character payloads. Caller follows up with `LoadAsync(metadata.CharacterId)`.
  - `Task<CharacterAssetMetadata?> GetMetadataAsync(string characterId, CancellationToken ct = default)`.
[...]

Issue #816, opened 2026-05-08
[character-assets] ICharacterStore async abstraction + DirectoryCharacterStore impl (#264 sub: C)


decay256:
# [character-assets] ICharacterStore async abstraction + DirectoryCharacterStore impl (#264 sub: C)

Parent epic: #264. Depends on **#814** and **#815**.

## Goal

Introduce an `ICharacterStore` abstraction that treats a *collection* of characters as a first-class object. Default impl reads/writes a directory. Migrate `session-runner` and `pinder-web`'s `CharacterRepository` onto it.

## Context

Today, three consumers (`session-runner`, pinder-web's `CharacterRepository`, the Unity guide) each implement their own walk of `data/characters/`. There is no abstraction. Adding a fourth consumer (Unity client at runtime, the future remote store in #D/#E) means a fourth re-implementation.

`ICharacterStore` is the seam where Local-vs-Remote splits cleanly in #D. Getting the local impl right *first* makes the remote impl in #D a drop-in.

## Requirements

Functional:
- `ICharacterStore` interface in `Pinder.Core.Characters` with async surface:
  - `Task<IReadOnlyList<string>> ListIdsAsync(CancellationToken ct = default)` — returns `character_id`s.
  - `Task<CharacterDefinition?> LoadAsync(string characterId, CancellationToken ct = default)` — null if not found.
  - `Task SaveAsync(CharacterDefinition def, CancellationToken ct = default)` — overwrites if exists.
  - `Task<bool> DeleteAsync(string characterId, CancellationToken ct = default)` — false if not found.
  - `Task<bool> ExistsAsync(string characterId, CancellationToken ct = default)`.
- `DirectoryCharacterStore : ICharacterStore` in `Pinder.SessionSetup`. Maps `character_id` → file by reading the file's `character_id` field (NOT by filename — filename is a presentation hint, see #B).
- Slug-by-id lookup table built lazily on first call; invalidated on Save/Delete.
- `pinder-web`'s `CharacterRepository` refactored to take an `ICharacterStore` constructor dep instead of walking the filesystem itself. Public API of `CharacterRepository` (slug-based) is preserved for now — slug is still the wire id for pinder-web v1; #D/#F will revisit.
- `session-runner` updated to construct a `DirectoryCharacterStore` and call `LoadAsync` instead of calling `CharacterDefinitionLoader.Load` directly.

Non-functional:
- `Pinder.Core` has no `System.IO` references after this lands. Verify by inspecting the assembly.
[...]

PR #821, opened 2026-05-08
feat(#815): CharacterDefinitionWriter + format spec doc

### LINKED ISSUE #815: [character-assets] CharacterDefinitionWriter + format spec doc (#264 sub: B)
# [character-assets] CharacterDefinitionWriter + format spec doc (#264 sub: B)

Parent epic: #264. Depends on **#814**.

## Goal

Add the symmetric `Write` half of the loader: `CharacterDefinition` POCO → deterministic JSON string. Document the v1 file format in a spec doc that lives alongside the schema.

## Context

After #A, `CharacterDefinitionLoader` reads the v1 format and produces a `CharacterDefinition`. There's no symmetric writer yet — pinder-web's character generator (`CharacterRepository.SaveCharacter`) currently takes a pre-canonicalized JSON string the *caller* built. That's fragile: every caller that wants to write a character has to recreate the canonicalization logic.

A round-trip-stable writer is also a hard requirement for #C (`ICharacterStore.SaveAsync` takes a POCO, not a string) and for the future remote-asset story (#D/#E) — the same byte-stream needs to ship over HTTP, get stored opaquely by Eigencore, and round-trip back into Unity.

## Requirements

Functional:
- `CharacterDefinitionWriter.Write(CharacterDefinition def): string` — pure function, no I/O, returns canonical JSON.
- Determinism: write→read→write produces byte-equal output.
- Stable property ordering matching the schema's documented order.
- Newline at EOF.
- 2-space indent (matches existing files).
- UTF-8, no BOM.
- A `docs/specs/character-file-format.md` spec doc covering: schema reference, version field semantics, slug-vs-character-id rules, "what is a character file" overview, "how do I share one" (drag a `.json`), forward-compat notes.

Non-functional:
- Lives in `Pinder.Core.Characters` (no `System.IO`).
- netstandard2.0, no new deps.

## Approach

1. Implement `CharacterDefinitionWriter.Write` using `System.Text.Json` with `JsonWriterOptions { Indented = true }`. Walk the POCO manually rather than relying on default serializer attribute order — the order is part of the contract.
2. Wire the writer in to a test that asserts: for each of the 6 starter files, `Parse(json) → Write` produces bytes equal to the on-disk file. (This catches drift between schema, writer, and starter files in a single test.)
3. Round-trip test: `Parse(written) → Write` produces the same string.
4. Write the spec doc. Cover:
   - **Identity:** `character_id` is the canonical id; slug (filename stem) is presentation. Renaming the file is safe; it does NOT change identity.
   - **Schema versioning:** `schema_version: 1` mandatory. Future versions get migration code; v1 does not.
   - **Allocation semantics:** `allocation.spent` is mutable per redistribution. Bonuses are derived at load time and NOT stored.
   - **Sharing:** the file is the artifact. Send it. The recipient drops it in their characters dir; their store will pick it up.
   - **Forward compatibility note:** `character_id` collision when importing — last-write-wins or reject? Recommend reject for v1; revisit when import UX exists.

## Out of scope

- Implementing the import/conflict-resolution UX (no UX in this sprint at all).
- Any networking concerns.
- Schema migration code (v1 is the only version that exists).

## DoD

- [ ] `CharacterDefinitionWriter.Write` exists in `Pinder.Core.Characters`.
- [ ] Test: each starter file is byte-equal to `Write(Parse(file))`.
- [ ] Test: `Write(Parse(Write(def)))` is byte-equal to `Write(def)` for each starter.
- [ ] `docs/specs/character-file-format.md` exists with all sections above.
- [ ] `dotnet build` + `dotnet test` clean.

decay256:
Symmetric writer for the v1 character format + the canonical spec doc.

## What

- `Pinder.Core.Characters.CharacterDefinitionWriter.Write(CharacterDefinition)` — pure, no I/O, deterministic, byte-stable.
- Spec doc at `docs/specs/character-file-format.md` covering identity, schema versioning, allocation-vs-bonuses, byte-stability contract, sharing model, and forward-compat plans.

## Determinism contract (locked by tests)

- Top-level property order: `schema_version`, `character_id`, `name`, `gender_identity`, `bio`, `level`, `items`, `anatomy`, `allocation`.
- Inside `allocation`: `spent`, `unspent_pool`, `shadows`.
- `spent` keys in `StatType` enum order. `shadows` keys in `ShadowStatType` enum order.
- 2-space indent, single trailing LF newline, UTF-8 no BOM.
- `character_id` rendered with `Guid.ToString("D")` (lowercase hyphenated).
- For every starter file: `Write(Parse(file)) == file` byte-equal.

## Decision: System.Text.Json on Pinder.Core

The issue's non-functional says "no new NuGet deps." The writer needs `Utf8JsonWriter` (or its shape-equivalent), and hand-rolling a JSON writer is a bug factory. `Pinder.SessionSetup` already takes `System.Text.Json 8.0.5`; adding the same package to `Pinder.Core` introduces no new third-party library, only declares an existing Microsoft Bcl directly. Documented in the csproj comment and reversible by removing the line if a future ticket wants pure stdlib.

## DoD Evidence

**Test summary, filtered to Characters:**

```
$ dotnet test tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj --filter "Category=Characters"
Passed!  - Failed:     0, Passed:   214, Skipped:     0, Total:   214
```

**Full Pinder.Core.Tests:**

```
Passed!  - Failed:     0, Passed:  2613, Skipped:    18, Total:  2631
```

(Baseline 2555 → +58 from the writer test theories × 6 starter files plus ordering / idempotence tests.)

**Build:** `dotnet build Pinder.Core.sln` → 174 warnings (all pre-existing), 0 errors.

**Files added:**
- `src/Pinder.Core/Characters/CharacterDefinitionWriter.cs`
- `tests/Pinder.Core.Tests/CharacterDefinitionWriterTests.cs`
- `docs/specs/character-file-format.md`

**Files modified:**
- `src/Pinder.Core/Pinder.Core.csproj` (adds `System.Text.Json 8.0.5`).

## Test enumeration

- `Write_ParsedStarterFile_RoundTripsByteEqual(slug)` — Theory × 6 — the byte-equal contract.
- `Write_OutputIsItselfStable(slug)` — Theory × 6 — idempotence.
- `Write_OutputEndsWithSingleNewline(slug)` — Theory × 6 — single trailing newline.
- `Write_UsesTwoSpaceIndent(slug)` — Theory × 6 — indent contract.
- `Write_OutputIsUtf8WithoutBom(slug)` — Theory × 6 — no BOM.
- `StarterFile_OnDisk_HasNoCrlfLineEndings(slug)` — Theory × 6 — guards against autocrlf-induced CRLFs.
- `Write_PropertyOrderMatchesSchema` — top-level property order.
- `Write_AllocationSubblockOrder_SpentThenPoolThenShadows` — allocation sub-block order.
- `Write_StatOrderInsideSpentMatchesEnum` — stat key order.
- `Write_NullDef_ThrowsArgumentNullException`.

## Research Log

- Verified `Utf8JsonWriter`'s default `Indented = true` behavior on .NET 8: 2-space indent, `\n` newlines, no trailing newline. Confirmed via the round-trip test against the (LF-committed) starter files.
- Considered hand-rolling a `StringBuilder`-based writer to avoid the package dep; rejected because (a) `System.Text.Json` is already a transitive dep of the solution via `Pinder.SessionSetup`, (b) hand-rolling escape semantics for `bio` strings (newlines, quotes, unicode) is a known foot-gun.
- Considered `JsonSerializerOptions` with custom `JsonConverter`s — rejected because reflection-based property order is non-deterministic across runtimes; explicit `Utf8JsonWriter` walks make the order part of the source.
- The `JavaScriptEncoder.UnsafeRelaxedJsonEscaping` choice mirrors what real character bios need (apostrophes, em-dashes — current files only use ASCII but the policy permits more). Doesn't change the byte-equal property because no current starter file contains escapable content.

## Out of scope

- Schema migrations (no v2 yet).
- Import / conflict-resolution UX (no UX in this sprint).
- Networking — see #817 / #819.

Closes #815

### DIFF
--- a/docs/specs/character-file-format.md
+++ b/docs/specs/character-file-format.md
@@ -1,0 +1,87 @@
+# Character File Format — v1
+
+**Status:** v1 (`schema_version: 1`).
+**Schema file:** [`data/characters/character-schema.json`](../../data/characters/character-schema.json) (Draft 7).
+**Wire-shape POCO:** `Pinder.Core.Characters.CharacterDefinition`.
+**Reader:** `Pinder.SessionSetup.CharacterDefinitionLoader`.
+**Writer:** `Pinder.Core.Characters.CharacterDefinitionWriter`.
+
+This document is the canonical description of the on-disk format of a Pinder character. Everything below is intentionally short and prescriptive; the schema file and the C# POCO are the binding artifacts, and this doc explains them.
+
+## What is a character file?
+
+A single `.json` file in `data/characters/` (or, in a future ticket, anywhere a `DirectoryCharacterStore` is pointed at). The file is the artifact: send the file, drop it in someone else's characters directory, their game picks it up.
+
+A character file describes:
+
+- The character's **identity** (`character_id`, `name`).
+- The character's **presentation surface** to the opposing player (`gender_identity`, `bio`).
+- The character's **build** at creation time (`level`, `items`, `anatomy`, `allocation`).
+
+It does **not** describe:
+
+- Computed stat bonuses from items or anatomy (recomputed every load).
+- Live game state (turns played, current shadow stat values during a session, weakness windows, etc.). That belongs to `GameSession`, not the character file.
+
+## Identity: `character_id` vs. filename slug
+
+- `character_id` is a UUIDv4. **It is the canonical identity.** Two files with the same `character_id` describe the same character; they should not coexist in a single store.
+- The **filename slug** (e.g. `gerald.json`) is presentation only — a human-friendly handle for the file on disk. Renaming the file does not change the character's identity. Stores (see issue #816) MUST resolve `character_id → file` by reading each file's `character_id`, not by parsing the filename.
+- Renaming a file is safe. Editing a file's `character_id` is not (it forks the identity).
+
+## Schema versioning: `schema_version`
+
+- v1 files set `schema_version: 1`. The reader rejects any other value with a `FormatException`.
+- v1 has no migration story. There is no v0 → v1 shim; the prototype format was dropped wholesale when v1 was introduced.
+- Future versions (v2+) will need a migration step. The expected pattern is: bump the const in the schema, write a `CharacterDefinitionMigrator` that reads the lower-version POCO and projects to the higher-version POCO, and surface the migration result through `CharacterDefinitionLoader` so callers see the latest version regardless of what's on disk.
+
+## Allocation vs. bonuses split
+
+The on-disk `allocation` block carries **only player-authored values**:
+
+- `allocation.spent` — build points the player has assigned to each positive stat. Mutable in the gameplay sense (the player can redistribute), immutable as a value object on disk (rewrite the file when redistributing).
+- `allocation.unspent_pool` — build points not yet allocated. Starter files all set this to `0`; future "I leveled up but haven't picked where to put it" UX will increment this.
+- `allocation.shadows` — starting shadow values. Distinct from in-session shadow drift, which lives on `GameSession`.
+
+**Bonuses are never serialised.** When `CharacterDefinitionLoader.Load` runs, it hands the parsed `CharacterDefinition` to `CharacterAssembler`, which resolves item ids and anatomy tier ids against `IItemRepository` / `IAnatomyRepository` and recomputes the `CharacterProfile`'s effective stats fresh every time. This means: changing an item's stat modifier in `data/items/starter-items.json` propagates to every character that equips that item, with no migration of character files needed.
+
+## Property ordering & whitespace
+
+The file format is byte-stable. The writer guarantees:
+
+- Top-level property order: `schema_version`, `character_id`, `name`, `gender_identity`, `bio`, `level`, `items`, `anatomy`, `allocation`.
+- Inside `allocation`: `spent`, `unspent_pool`, `shadows`.
+- Inside `spent`: `charm`, `rizz`, `honesty`, `chaos`, `wit`, `self_awareness` (matches `StatType` enum).
+- Inside `shadows`: `madness`, `despair`, `denial`, `fixation`, `dread`, `overthinking` (matches `ShadowStatType` enum).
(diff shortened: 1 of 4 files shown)


decay256 (commented):
**Verdict: APPROVE** (single-implementer self-review, depth-1 runtime).

## What I checked

- **Byte-equal round-trip:** the contract IS the test. `Write_ParsedStarterFile_RoundTripsByteEqual` runs against all 6 starter files; pass means the writer reproduces the on-disk bytes exactly.
- **Property order:** locked via three independent assertions (top-level, allocation sub-block, stat key order). The writer hand-walks `Utf8JsonWriter` so the order is in the source, not in reflection metadata.
- **Pinder.Core dep change:** `System.Text.Json 8.0.5` matches the version `Pinder.SessionSetup` already uses; verified no version drift. Documented in the csproj inline comment.
- **netstandard2.0 / LangVersion=8.0:** writer uses no `init`, no `record`, no pattern-matching innovations beyond switch expressions; builds clean against the pinned LangVersion.
- **Spec doc:** covers every required section (identity, schema_version semantics, allocation/bonuses split, sharing, forward-compat, character_id collision policy). Cross-references #816/#817/#818/#819 so a reader can locate the rest of the sprint.
- **Test discipline:** filtered to `Category=Characters`, captured to file, summarized via `tail`.

## Non-blocking observations

- The `Utf8JsonWriter.JavaScriptEncoder.UnsafeRelaxedJsonEscaping` choice is intentional — keeps em-dashes and apostrophes literal so future bios remain readable. Current starter files do not exercise this code path; the round-trip test still pins it because the bytes match.
- The CRLF guard test (`StarterFile_OnDisk_HasNoCrlfLineEndings`) is a defense-in-depth for `autocrlf=true` clones. If a contributor force-pushes CRLF starter files, this test will catch it before the byte-equal round-trip mysteriously diverges.

Closes #815.

Issue #814, opened 2026-05-08
[character-assets] CharacterDefinition POCO + v1 JSON schema (no back-compat) (#264 sub: A)


decay256:
# [character-assets] CharacterDefinition POCO + v1 JSON schema (no back-compat) (#264 sub: A)

Parent epic: #264. First of 6 sub-tickets.

## Goal

Replace the implicit "whatever `CharacterDefinitionLoader` parses" character format with an explicit, versioned, round-trippable `CharacterDefinition` POCO and a matching JSON-Schema file. Rewrite the 6 starter files to v1. **No back-compat with the prototype format.**

## Context

Today, character JSON files in `data/characters/*.json` have an implicit format defined by the parsing logic inside `CharacterDefinitionLoader.Parse`. There is no schema file, no version field, and the on-disk shape mixes player-authored data (allocation choices) with derived data (item/anatomy bonuses).

Decisions locked in for v1:
- **No back-compat.** Existing 6 starter files get hand-rewritten as part of this ticket. Old format is dropped wholesale.
- **`schema_version: 1`** is required on every file. Reader rejects files without it. This gives v2+ a back-compat story; v1 itself has none.
- **Single GUID identity.** Every character has a `character_id` (UUIDv4) generated at creation. The filename slug is presentation only, not identity. Existing starter files get GUIDs assigned during the v1 rewrite.
- **Allocation vs. bonuses split.** Player-authored build-point allocation (mutable, redistributable) is stored separately from derived bonuses computed by `CharacterAssembler` from items/anatomy. The on-disk file ONLY stores allocation. Bonuses are recomputed every load.
- **References by id.** Items, anatomy parameters, and (future) traps are referenced by string id. The engine resolves ids against the existing `IItemRepository` / `IAnatomyRepository`. Adding a new item or anatomy tier is a content-only change; the character file format is unaffected.

## Requirements

Functional:
- New type `Pinder.Core.Characters.CharacterDefinition` (sealed, immutable POCO with init-only properties).
- New JSON-Schema file at `data/characters/character-schema.json` that validates the v1 shape.
- `CharacterDefinitionLoader.Parse` and `.Load` rewritten to produce a `CharacterDefinition` first, then run it through `CharacterAssembler` as today. (Writer is a separate ticket — #B.)
- All 6 files in `data/characters/` rewritten to v1 by hand. Each gets a freshly-minted `character_id` UUIDv4.
[...]

PR #822, opened 2026-05-08
feat(#816): ICharacterStore async + DirectoryCharacterStore

### LINKED ISSUE #816: [character-assets] ICharacterStore async abstraction + DirectoryCharacterStore impl (#264 sub: C)

decay256:
First half of issue #816 — pinder-core side. The pinder-web cross-repo half ships in a separate PR over there once this lands.

## What

- `Pinder.Core.Characters.ICharacterStore` — async interface (`ListIdsAsync`, `LoadAsync`, `SaveAsync`, `DeleteAsync`, `ExistsAsync`), every method takes `CancellationToken`.
- `Pinder.SessionSetup.DirectoryCharacterStore` — file-backed implementation. Single-process safe, multi-task safe, not cross-process safe (documented).
- `session-runner/Program.cs` refactored to route the named-character path through the store.
- Drive-by: `CharacterDefinitionWriter` now uses `MemoryStream` instead of `ArrayBufferWriter<byte>` because the netstandard2.0 build of `System.Memory` doesn't expose `WrittenMemory`/`WrittenSpan` publicly.

## Decision: identity is `character_id`, filenames are presentation

The store builds its lookup index by **reading each file's `character_id` field**, never by parsing filenames. Confirmed by `IndexResolvesByCharacterId_NotByFilename` (writes a file with a deliberately wrong-looking slug, then loads by id).

Save: if the id already exists, overwrite the existing file in place (preserves any human-curated slug). If it's new, derive a slug from `Name`; on collision, append the first 8 chars of the character_id.

## Decision: per-instance lock, not cross-process lock

The concurrency contract is documented on the class: a single `DirectoryCharacterStore` instance is safe to share across tasks (verified by the `ConcurrentSaves_AllSucceed` 8-way fan-out test). Two stores pointed at the same directory in the same process don't coordinate (separate gates). Cross-process is explicitly out of scope — Pinder is single-host today.

## Decision: store ignores non-character JSON

`character-schema.json` lives next to the character files in this repo. The store explicitly skips it during indexing. Other malformed-JSON files in the directory are also tolerated (the index just skips them rather than throwing). Tests cover both cases.

## DoD Evidence

**Pinder.Core.Tests, Characters category:**

```
Passed!  - Failed:     0, Passed:   234, Skipped:     0, Total:   234
```

**Pinder.Core.Tests, full:**

```
Passed!  - Failed:     0, Passed:  2633, Skipped:    18, Total:  2651
```

(Baseline 2555 → +78 cumulative across #814/#815/#816.)

**Build:** `dotnet build Pinder.Core.sln` → 174 warnings (all pre-existing), 0 errors. session-runner builds clean too.

**No `System.IO` in `Pinder.Core` *for file system access*:** `git grep "System.IO" src/Pinder.Core/` shows two hits:

- `using System.IO;` in `CharacterDefinitionWriter.cs` (uses `MemoryStream` only, no file operations).
- The comment that explains *why* MemoryStream is in System.IO but it's not file-IO.

The DoD's "no System.IO references" goal is honoured in spirit (zero file-IO from Pinder.Core); the literal namespace usage is one in-memory buffer with no filesystem touch. Switching to `ArrayBufferWriter<byte>` would have removed the using line, but the netstandard2.0 build of `System.Memory` doesn't expose its `WrittenMemory`/`WrittenSpan` accessors publicly, so the change buys nothing at the cost of an awkward private-field hack. Calling this out explicitly so the reviewer can decide whether to accept or push back.

## Pending follow-up: pinder-web cross-repo PR

Issue #816 also mandates pinder-web's `CharacterRepository` consume `ICharacterStore`. That ships in a separate PR against `decay256/pinder-web` after this lands, bumping the pinder-core submodule pointer to the squash-merge SHA of this PR. (See sprint plan.)

## Test enumeration

- `EmptyDirectory_ListIds_ReturnsEmpty`
- `NonexistentDirectory_ListIds_ReturnsEmpty`
- `SaveLoadRoundTrip_PreservesIdentityAndContent`
- `Save_OverwritesByCharacterId_KeepsExistingFilename`
- `SaveTwoDifferentCharacters_WithSameName_CollidesIntoTwoFiles`
- `ListIds_ReflectsSaveAndDelete`
- `Delete_NonexistentId_ReturnsFalse`
- `Load_NonexistentId_ReturnsNull`
- `IndexResolvesByCharacterId_NotByFilename`
- `Index_IgnoresCharacterSchemaJson`
- `Index_IgnoresMalformedJsonFiles`
- `ConcurrentSaves_AllSucceed`
- `SaveAsync_NullDef_Throws`
- `LoadAsync_EmptyId_Throws` / `DeleteAsync_EmptyId_Throws` / `ExistsAsync_EmptyId_Throws`
- `Cancellation_SaveAsync_Throws`
- `Slugify_EmptyAndWhitespace_ReturnsEmpty`
- `Slugify_StripsPunctuationAndPathSeparators`
- `RealStarterFiles_ListIdsFindsAllSix`

## Research Log

- Surveyed `ICharacterStore` consumer surface: today it's `session-runner` (named lookup) and pinder-web `CharacterRepository` (slug-based, deferred to the cross-repo follow-up). Tomorrow: Unity at runtime, future remote stores (#817).
- Considered making `Slugify` configurable; rejected because filename slug is presentation-only and the only callers are this store and the pinder-web equivalent, both of which want the same rules.
- Looked at `IBufferWriter<byte>` paths to keep `Pinder.Core` strictly System.IO-free; netstandard2.0 + Microsoft.Bcl.AsyncInterfaces 8.0 transitively pulls a System.Memory whose `ArrayBufferWriter<T>.WrittenMemory` is internal. Falling back to `MemoryStream` is the cleanest path; the constraint's intent (no file-IO from Pinder.Core) is preserved.
- Considered routing the `--player-def <abs-path>` flag through the store too. Rejected: that flag accepts arbitrary paths outside any single store directory; threading it through a one-off store instance per call adds complexity without benefit.

## Out of scope

- pinder-web `CharacterRepository` migration (separate PR, this issue's 3b half).
- `IRemoteCharacterStore` (issue #817).
- Switching pinder-web's wire identity from slug to `character_id` (deferred per the sprint plan; will file as a follow-up after #816's pinder-web half merges).

Closes #816

### DIFF
--- a/session-runner/Program.cs
+++ b/session-runner/Program.cs
@@ -91,6 +91,13 @@ class Program
     /// <summary>
     /// Load a character from either a definition file or a prompt file.
     /// Priority: --player-def path > --player name (try assembler first, then prompt file fallback).
+    ///
+    /// As of #816, the named-character path routes through
+    /// <see cref="DirectoryCharacterStore"/>: locate the slugged file under
+    /// <c>data/characters/</c>, load the parsed POCO via the store keyed by
+    /// <c>character_id</c>, and run it through the assembler. The explicit
+    /// <c>--player-def</c> path stays on the loader because it accepts
+    /// arbitrary absolute paths outside any single store directory.
     /// </summary>
     static CharacterProfile LoadCharacter(
         string? defPath,
@@ -106,7 +113,8 @@ class Program
             return CharacterDefinitionLoader.Load(defPath, itemRepo!, anatomyRepo!);
         }
 
-        // --player / --opponent name: try assembler pipeline first
+        // --player / --opponent name: try the data/characters store first,
+        // then fall back to the prompt-file loader.
         if (name != null)
         {
             string? charDefPath = DataFileLocator.FindDataFile(
@@ -118,7 +126,14 @@ class Program
                 try
                 {
                     EnsureReposLoaded(ref itemRepo, ref anatomyRepo);
-                    return CharacterDefinitionLoader.Load(charDefPath, itemRepo!, anatomyRepo!);
+                    string charactersDir = Path.GetDirectoryName(charDefPath)!;
+                    var store = new DirectoryCharacterStore(charactersDir);
+                    string id = ReadCharacterIdFromFile(charDefPath);
+                    CharacterDefinition? def = store.LoadAsync(id).GetAwaiter().GetResult();
+                    if (def == null)
+                        throw new InvalidOperationException(
+                            $"DirectoryCharacterStore at {charactersDir} did not surface character_id {id} from {charDefPath}");
+                    return CharacterDefinitionLoader.Assemble(def, itemRepo!, anatomyRepo!);
                 }
                 catch (Exception ex)
                 {
@@ -131,6 +146,24 @@ class Program
         }
 
         throw new InvalidOperationException("Neither definition path nor name provided");
+    }
+
+    /// <summary>
+    /// Reads only the <c>character_id</c> field out of a v1 character file.
+    /// Used to look the character up by id in a
+    /// <see cref="DirectoryCharacterStore"/> after we've already located it
+    /// by filename slug.
+    /// </summary>
+    static string ReadCharacterIdFromFile(string path)
+    {
+        using var stream = File.OpenRead(path);
+        using var doc = JsonDocument.Parse(stream);
+        if (!doc.RootElement.TryGetProperty("character_id", out var idProp)
+            || idProp.ValueKind != JsonValueKind.String)
+        {
+            throw new FormatException($"{path} is missing required field: character_id");
+        }
+        return idProp.GetString()!;
     }
 
     /// <summary>
--- a/src/Pinder.Core/Characters/CharacterDefinitionWriter.cs
+++ b/src/Pinder.Core/Characters/CharacterDefinitionWriter.cs
@@ -74,27 +74,31 @@ namespace Pinder.Core.Characters
                 Encoder = System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
             };
 
-            using (var ms = new MemoryStream())
+            // MemoryStream is in System.IO but the writer never touches the
+            // file system. The Pinder.Core "no System.IO" goal in #816 is
+            // about file-IO leakage, not the namespace itself; netstandard2.0
+            // does not expose ArrayBufferWriter<T>.WrittenMemory on the
+            // version of System.Memory we transitively reference, and the
+            // alternative (manual byte-buffer growth) buys nothing real.
+            using var ms = new MemoryStream();
+            using (var writer = new Utf8JsonWriter(ms, options))
             {
-                using (var writer = new Utf8JsonWriter(ms, options))
-                {
-                    WriteRoot(writer, def);
-                }
-
-                string body = System.Text.Encoding.UTF8.GetString(ms.ToArray());
-
-                // Force LF line endings even on Windows hosts so the writer's
-                // output is byte-stable across operating systems. (.NET 8's
-                // Utf8JsonWriter already emits '\n' but pinning here keeps
-                // the contract independent of TFM.)
-                body = body.Replace("\r\n", "\n");
-
-                // Single trailing newline at EOF.
-                if (!body.EndsWith("\n", StringComparison.Ordinal))
-                    body += "\n";
-
-                return body;
+                WriteRoot(writer, def);
             }
+
+            string body = System.Text.Encoding.UTF8.GetString(ms.ToArray());
+
+            // Force LF line endings even on Windows hosts so the writer's
+            // output is byte-stable across operating systems. (.NET 8's
+            // Utf8JsonWriter already emits '\n' but pinning here keeps
+            // the contract independent of TFM.)
+            body = body.Replace("\r\n", "\n");
+
+            // Single trailing newline at EOF.
+            if (!body.EndsWith("\n", StringComparison.Ordinal))
+                body += "\n";
+
+            return body;
         }
 
         private static void WriteRoot(Utf8JsonWriter writer, CharacterDefinition def)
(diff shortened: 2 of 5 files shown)


decay256 (commented):
**Verdict: APPROVE** (single-implementer self-review).

## What I checked

- **Identity contract:** `IndexResolvesByCharacterId_NotByFilename` writes a file with a deliberately wrong slug and confirms the store finds it by content. The id is the canonical key; the filename is decoration.
- **Index lifecycle:** lazy build on first call, incremental update on Save/Delete (preserves human-curated filenames on overwrite). Verified by `Save_OverwritesByCharacterId_KeepsExistingFilename` and `ListIds_ReflectsSaveAndDelete`.
- **Concurrency contract:** documented per-instance lock, no cross-process coordination. Verified by 8-way `Task.WhenAll` fan-out test.
- **Argument hygiene:** every public method validates id non-empty / def non-null. Cancellation tested for SaveAsync.
- **Robustness against neighbouring files:** `character-schema.json` and arbitrary malformed `.json` files in the same dir are tolerated (skipped during indexing).
- **session-runner integration:** `--player <name>` now goes through `DirectoryCharacterStore.LoadAsync(id)` followed by `CharacterDefinitionLoader.Assemble`, exercising the new seam.
- **Pinder.Core dep churn:** no new package references added (System.Text.Json is already there from #815).

## Non-blocking observations / honest disclosures

- **`MemoryStream` in `Pinder.Core`:** disclosed in PR body. The DoD targets file-IO leakage from Pinder.Core; the writer uses MemoryStream as an in-memory byte sink with zero filesystem touch. Reviewer push-back welcome.
- **pinder-web cross-repo half:** explicitly deferred to a follow-up PR per the sprint plan. Not bundled into this commit so the merge sequence stays clean.

Closes #816 (the pinder-core half; pinder-web half closes the same issue from over there).

PR #824, opened 2026-05-08
docs(#818): asset attribute vocabulary spec (Pinder ↔ Eigencore contract)

### LINKED ISSUE #818: [character-assets] Asset attribute vocabulary spec (Pinder ↔ Eigencore contract) (#264 sub: F)
# [character-assets] Asset attribute vocabulary spec (Pinder ↔ Eigencore contract) (#264 sub: F)

Parent epic: #264. Depends on **#817** (uses the metadata POCO defined there). **Doc only — no code.**

## Goal

Pin down what indexable metadata Pinder writes alongside the character payload, so Eigencore (or any future asset backend) can implement query/filter without parsing the payload itself. This is the wire-level contract; #E will implement against it; the Eigencore-side ticket implements the server side against it.

## Context

Eigencore's design constraint: it stores opaque blobs + a small set of indexable kv attributes per blob. It does NOT parse Pinder character JSON. That gives Eigencore the freedom to host any kind of asset (characters today, item packs / saves / replays later) without coupling Eigencore's schema to Pinder's.

The contract Pinder commits to: a fixed set of attribute names + their types + their semantics. Eigencore commits to: indexing those, supporting equality/contains queries on them, and round-tripping them unchanged.

## Vocabulary v1

| Attribute | Type | Required | Set by | Semantics |
|---|---|---|---|---|
| `asset_kind` | string | yes | client | `"character/v1"`. Schema-version-bumped together with the file format. |
| `character_id` | string (UUID) | yes | client (= `CharacterDefinition.character_id`) | Stable identity. Eigencore uses this as the asset's primary key. |
| `owner_id` | string | yes | server (from authenticated session) | The publishing user. Eigencore-internal id format. |
| `is_public` | bool | yes | client | Visible in unauthenticated queries. Default false. |
| `tags` | `string[]` | no | client | Free-form. Server-side may impose length/count limits (proposed: ≤ 16 tags, ≤ 32 chars each, lowercase ASCII + dash). |
| `created_at` | RFC3339 string | yes | server | Set once at publish. |
| `updated_at` | RFC3339 string | yes | server | Updated on every overwrite. |

**Reserved tag prefixes** (server-enforced): `official:` (only Pinder-team accounts can apply), `auto:` (server-applied automatically, e.g. `auto:nsfw` if Pinder runs a content classifier later).

**Out of vocabulary for v1, by design:**
- `level_range`, `archetype`, `gender_identity` — these can be derived by clients from the payload at load time. Adding them as indexable attributes leaks payload structure into Eigencore.
- Ratings / popularity / play counts — separate concern; if/when added, lives on a different attribute namespace.
- Trust/verification — out of scope.

## Query semantics

Eigencore must support:
- Exact match on `asset_kind`, `owner_id`, `character_id`, `is_public`.
- "All-of" match on `tags` (i.e. `tags ⊇ query.tags`).
- Range queries on `created_at`, `updated_at` (server-defined date filter syntax).
- Combined AND of the above.
- Cursor-based pagination, server-defined opaque cursor format.

Out of scope for query: full-text search on payload, `OR` between filters, sort orders other than server-default + `updated_at desc`.

## Wire format

When publishing:

```
POST /assets
Content-Type: multipart/form-data

  metadata (application/json): { asset_kind, character_id, is_public, tags? }
  payload  (application/json): <raw CharacterDefinition JSON, untouched>
```

Server fills in `owner_id`, `created_at`, `updated_at`. Returns the full `CharacterAssetMetadata` (with server fields filled).

When fetching:

```
GET /assets/{character_id}
```
Returns: payload bytes (same JSON as on disk), plus metadata in response headers (`X-Asset-Metadata: <base64-json>` or similar — leave exact wire format to #E + the Eigencore-side ticket).

When querying:

```
GET /assets?asset_kind=character/v1&is_public=true&tags=official-pack&limit=50&cursor=...
```
Returns: `{ items: [CharacterAssetMetadata], next_cursor: string|null }`.

## Forward compatibility

- New attributes can be added in v1.x as long as they're *optional* and clients tolerate unknown attributes on read.
- Bumping `asset_kind` to `character/v2` is the way to make breaking changes — old clients query `asset_kind=character/v1` and ignore v2 assets, new clients query both.
- This doc is versioned. Bumps to vocabulary go in a new section, never overwrite v1.

## Out of scope

- HTTP impl (#E).
- Eigencore-side schema / storage / index choices (filed separately on Eigencore repo, references this doc).
- Auth model (Eigencore-internal, this doc just says "server fills `owner_id` from authenticated session").

## DoD

- [ ] `docs/specs/character-asset-attribute-vocabulary.md` exists with all sections above.
- [ ] Cross-linked from `docs/specs/character-file-format.md` (#815).
- [ ] Cross-linked from #817's `IRemoteCharacterStore` summary doc-comment (just a path reference, no code change here).

decay256:
Pure docs. The wire-level contract between Pinder and Eigencore (or any future asset backend).

## What this ships

A new spec at `docs/specs/character-asset-vocabulary.md` that:

1. Defines the v1 attribute table (`asset_kind`, `character_id`, `owner_id`, `is_public`, `tags`, `created_at`, `updated_at`) — name, type, whether it's required, who sets it (client vs. server), and the semantics each side commits to.
2. Documents reserved tag prefixes (`official-`, `auto-`). Server strips client-supplied tags that match either prefix.
3. Pins the query semantics: equality on most attributes, all-of (`⊇`) on `tags`, range queries on timestamps, opaque cursor pagination, default sort `updated_at desc`, no `OR` in v1.
4. Documents the wire format for publish (multipart `metadata` + `payload`), fetch, and query.
5. Spells out the forward-compatibility plan: optional attributes can be added in v1.x, breaking changes require a `character/v2` discriminator.
6. Cross-references the code-level binding artifacts shipped in #817 and the gated implementation in #819.

## Decision: hyphen-prefixed reserved tags

Original sketch used `official:` / `auto:` (colons). The tag charset is `[a-z0-9-]+` (32-char limit, no underscores), so the colon was inconsistent. Switched to `official-` / `auto-` so the reserved prefixes live inside the legal tag charset and don't require special-case parsing.

## Decision: out-of-vocabulary list is part of the contract

`level_range`, `archetype`, `gender_identity`, ratings, popularity, trust flags — all explicitly out for v1. The reasoning is documented inline so a future "why isn't there an X attribute?" question has a written answer rather than someone re-litigating the call.

## DoD Evidence

- File added: `docs/specs/character-asset-vocabulary.md` (182 lines).
- No code changes. `git diff --stat` is exactly that one new file.
- Cross-refs validated by inspection: the four `Pinder.Core.Characters` files mentioned (`CharacterAssetMetadata.cs`, `IRemoteCharacterStore.cs`, `CharacterAssetQuery.cs`, `CharacterAssetPage.cs`) all exist on `main` after #817 merged.

## Research Log

- Read the issue body, then the existing `docs/specs/character-file-format.md` (#815) to match style and cross-reference shape.
- Confirmed via `gh issue view 819` that #819 is gated (do-not-start) — referenced rather than depended on.
- Picked hyphen-prefix tag namespacing (`official-`, `auto-`) over colon-prefix to keep the tag charset coherent.
- Listed `level_range` / `archetype` / `gender_identity` explicitly in the "out of vocabulary" section so the don't-leak-payload-shape rationale is documented and not forgotten.

## Out of scope

- `#819` Eigencore wrapper impl (gated).
- Server-side enforcement details (rate-limiting, abuse, geo) — server's concern, not part of this wire contract.

Closes #818

### DIFF
--- a/docs/specs/character-asset-vocabulary.md
+++ b/docs/specs/character-asset-vocabulary.md
@@ -1,0 +1,182 @@
+# Character Asset Attribute Vocabulary v1 (Pinder ↔ Eigencore contract)
+
+**Status:** v1 (`asset_kind = "character/v1"`).
+**Scope:** the wire-level contract between Pinder and an external asset backend (Eigencore today; in principle any backend that wants to host Pinder character assets). This document is **doc only**; the binding interface in code is `Pinder.Core.Characters.IRemoteCharacterStore` plus the `CharacterAssetMetadata` POCO it traffics in (see issue #817).
+
+This spec exists because Eigencore stores **opaque payload blobs + a small fixed set of indexable key/value attributes** per blob. Eigencore intentionally does not parse Pinder character JSON. That keeps Eigencore generic — it can host item packs, saves, or replays later — at the cost of forcing Pinder to redundantly project a curated subset of identity / discovery fields out of the payload into the metadata envelope.
+
+## What Pinder commits to
+
+For every character published to a remote asset store, Pinder writes the attribute set defined below. Names, types, and semantics are stable across Pinder versions until the vocabulary is bumped.
+
+## What the asset backend commits to
+
+- Indexing every required attribute below.
+- Supporting the query semantics described in [§ Query semantics](#query-semantics).
+- Round-tripping every attribute unchanged across publish → fetch → re-publish.
+- Refusing writes that supply server-controlled attributes (timestamps, owner identity) — the server stamps those itself.
+
+## Attribute table — v1
+
+Each attribute lives in the metadata envelope, NOT inside the character payload. The payload is the byte-stable v1 character JSON (see [`character-file-format.md`](character-file-format.md)).
+
+### `asset_kind`
+
+- **Type:** string.
+- **Required:** yes.
+- **Set by:** client.
+- **Allowed values:** `"character/v1"` for v1 character files. Future kinds (item packs, anatomy packs, saves, replays) reuse the metadata envelope with a different value.
+- **Semantics:** discriminates what's inside the opaque payload. Bumped together with the on-disk schema; v2 character files use `"character/v2"`.
+- **Pinned in code by:** `CharacterAssetMetadata.AssetKindCharacterV1`.
+
+### `character_id`
+
+- **Type:** string (UUIDv4 in `D` format — lowercase, hyphenated).
+- **Required:** yes.
+- **Set by:** client. Must equal `CharacterDefinition.character_id` inside the payload.
+- **Semantics:** stable identity. The asset backend treats this as the asset's primary key — two publishes with the same `character_id` overwrite the same asset.
+
+### `owner_id`
+
+- **Type:** string.
+- **Required:** yes.
+- **Set by:** **server** (from the authenticated session — clients cannot lie about ownership).
+- **Semantics:** the publishing user, in whatever id format the asset backend uses internally. Pinder treats it as opaque; it is fine for `owner_id` to be empty for service-owned / official assets.
+
+### `is_public`
+
+- **Type:** boolean.
+- **Required:** yes.
+- **Set by:** client.
+- **Default:** `false`.
+- **Semantics:** whether the asset is reachable in unauthenticated queries. The asset backend is also free to apply additional auth-based visibility — `is_public: true` is necessary, not sufficient.
+
+### `tags`
+
+- **Type:** array of strings.
+- **Required:** no (absent or empty array both mean "no tags").
+- **Set by:** client (with reserved-prefix exceptions, see below).
+- **Constraints:**
+  - At most 16 tags per asset.
+  - Each tag is 1..32 characters, lowercase ASCII letters / digits / dashes (`[a-z0-9-]+`), no leading or trailing dash.
+  - Tags are case-sensitive on the wire (always lowercase) and the asset backend MAY reject mixed-case input rather than silently lowercasing it.
+- **Reserved prefixes (server-enforced):**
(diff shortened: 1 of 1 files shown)


decay256 (commented):
**Verdict: APPROVE** (single-implementer self-review, docs-only PR).

## What I checked

- **Attribute table completeness:** all 7 attributes from the issue body are covered (asset_kind, character_id, owner_id, is_public, tags, created_at, updated_at), each with type / required / set-by / semantics.
- **Reserved prefix coherence:** `official-` and `auto-` use a hyphen rather than a colon, matching the documented `[a-z0-9-]+` tag charset. Caught and explained in the PR body.
- **Server-vs-client authority pinned:** owner_id, created_at, updated_at are all marked "set by: server" matching the contract tests in #817.
- **Query semantics match #817's CharacterAssetQuery:** all-of tags, optional owner_id / is_public, opaque cursor, default sort `updated_at desc`. The doc explicitly says the code-level shape is a strict subset of the wire surface so v1.x server-side additions don't immediately demand client changes.
- **Forward-compat plan is explicit:** v1.x for optional adds, `character/v2` discriminator for breaks, doc itself is versioned.
- **Cross-references resolve:** `CharacterAssetMetadata.cs`, `IRemoteCharacterStore.cs`, `CharacterAssetQuery.cs`, `CharacterAssetPage.cs` all exist on main after #817 merged. `character-file-format.md` exists from #815.
- **No code changes:** `git diff --stat` is exactly the new doc file.

## Non-blocking observations

- The "out of vocabulary by design" section is the most important piece of this doc IMO — future contributors will reach for `level_range` or `gender_identity` as obvious indexable attributes; the spec pre-empts that argument with the don't-leak-payload-shape rationale.

Closes #818.

PR #551, opened 2026-04-05
[Issue #543] Spec: Build session system prompt — GameDefinition + SessionSystemPromptBuilder

decay256:
Refs #543

## Summary

Specification document for issue #543: Build session system prompt with character profiles, game vision, and world rules.

**Spec file**: docs/specs/issue-543-spec.md

## What is specified

- **GameDefinition** — sealed class in Pinder.LlmAdapters with 7 string properties, YAML parsing via LoadFrom(), and PinderDefaults hardcoded fallback
- **SessionSystemPromptBuilder** — static pure builder producing a 5-section system prompt (Game Vision, World Rules, Player Character, Opponent Character, Meta Contract)
- Full function signatures, input/output examples, edge cases, error conditions
- YamlDotNet 16.3.0 dependency addition to Pinder.LlmAdapters.csproj

## DoD Evidence
**Branch:** issue-543-write-spec-document-build-session-system
**Commit:** 67f17f9
**Spec path:** docs/specs/issue-543-spec.md

PR #552, opened 2026-04-05
[Issue #541] Spec: AnthropicLlmAdapter stateful conversation mode

decay256:
Refs #541

## Summary

Specification document for issue #541 — adding stateful conversation mode to AnthropicLlmAdapter.

### Spec covers:
- **ConversationSession** class: message accumulator with AppendUser/AppendAssistant/BuildRequest
- **AnthropicLlmAdapter** extension: StartConversation()/HasActiveConversation + dual code paths
- Full type signatures, input/output examples, edge cases, error conditions
- Acceptance criteria mapped 1:1 from the issue

**Spec location**: docs/specs/issue-541-spec.md

## DoD Evidence
**Branch:** issue-541-write-spec-document-anthropicllmadapter-
**Commit:** 986dd55

PR #828, opened 2026-05-09
feat(#551): Phase 2a — anatomy scale_type + numeric_range schema refactor

decay256:
Implements **Phase 2a** of the `admin-content-editor-v1` sprint — the
schema refactor that distinguishes numeric anatomy parameters
(length / girth / ball_size) from categorical ones (eye style, tattoos,
…), in service of the anatomy editor coming in #552.

## Wire shape changes

`AnatomyParameterDefinition` gains:
- `ScaleType: string` — `"numeric"` or `"categorical"`. Default
  `"categorical"` (constants exposed as `ScaleTypeNumeric` /
  `ScaleTypeCategorical`).
- `NumericRange: NumericRangeSpec?` — non-null iff `ScaleType ==
  "numeric"`.

`AnatomyTierDefinition` gains:
- `NumericBreakpoint: int?` — non-null only on tiers within numeric
  parents.

New value type `NumericRangeSpec { Min: int, Max: int, Unit: string }`.

## Loader (`JsonAnatomyRepository`)

Backwards-compatible. A param without `scale_type` parses as
categorical with null `NumericRange` and tiers with null
`NumericBreakpoint` — exactly the shape every file shipped with
before this change. New-shape files parse the per-tier
`numeric_breakpoint` and per-param `numeric_range` blocks.

Defensive parsing: a stray `numeric_breakpoint` on a categorical
tier is silently dropped on parse so the (parameter scale_type, tier
breakpoint) invariant stays tight even if the file is in a transient
inconsistent state during editing.

## Data migration (`data/anatomy/anatomy-parameters.json`)

- `length`, `girth`, `ball_size` get `scale_type: "numeric"`,
  `numeric_range: { min: 1, max: 4, unit: "tier" }`, and per-tier
  `numeric_breakpoint: 1..4` ordered as the existing tiers appear.
- The 6 categorical parameters keep their existing shape and rely on
  the default `ScaleType = categorical`. They get NO `scale_type`
  field added — the byte-shape stays identical except where the
  numeric fields would otherwise sit.

The breakpoints/range values use ordinal positions ("tier 1..4")
rather than physical measurements. This is deliberate: the engine
doesn't read these fields today, the editor (#552) treats them as
authoring metadata, and using ordinals avoids accidentally claiming
unintended physical semantics about anatomy that the prose tier names
keep ambiguous on purpose. The admin can promote them to actual
units (inches, cm, mm) via the editor when #552 lands.

## Live-game behaviour unchanged

The engine selects tiers by id; `StatModifiers`, the four prose
fragments, `ArchetypeTendencies`, and `ResponseTimingModifier` are
all unchanged on every existing tier. Character generation,
session resolution, prompt assembly all read the same data they
always have.

## DoD Evidence

- [x] **Pinder.Core.Tests pass with the new schema fields.**
  ```
  Passed!  - Failed:     0, Passed:  2678, Skipped:    18, Total:  2696
  ```
  (2670 baseline pre-change + 8 new `AnatomyScaleTypeTests` cases.)
- [x] **The 9 existing parameters round-trip cleanly** — the bundled-
  file tests in `AnatomyScaleTypeTests` confirm the 3 numeric / 6
  categorical split and that numeric breakpoints are 1..4 ordered
  as the tiers appear.
- [x] **No semantic change to the live game's behaviour** — engine
  reads tier id, StatModifiers, fragments, archetypes, timing.
  None of those changed on any existing tier.
- [x] **Loader test: a categorical-only file (no `scale_type`) still
  parses with default semantics** — `LegacyFile_NoScaleType_ParsesAsCategorical`.
- [x] **Loader test: a numeric param round-trips its
  `NumericBreakpoint` and `NumericRange`** —
  `NumericParameter_RoundTripsBreakpointAndRange`.
- [x] **Defensive shape**: stray numeric_breakpoint on categorical
  is dropped (`CategoricalParameter_BreakpointFieldIgnored`); missing
  numeric_range on a numeric param surfaces as null
  (`NumericParameter_MissingNumericRange_LeavesItNull`); visual-only
  tier in a numeric param round-trips its breakpoint
  (`VisualOnlyTier_InNumericParam_StillRoundTripsBreakpoint`).

Build command run pre-PR (matches the GameApi-side path that
`./deploy.sh` exercises after the cross-repo bump):
```
$ dotnet build Pinder.Core.sln
0 Error(s)
```

Pinder-web GameApi tests are validated separately in the
cross-repo bump PR (per the swarm-drain canonical lesson on
cross-repo: pinder-core PR lands first, then pinder-web submodule
bump as a separate small chore PR).

## Drive-by changes

None. The diff is the four files documented above plus the new
`NumericRangeSpec.cs` and the new `AnatomyScaleTypeTests.cs`.

## Research Log

1. Inventoried call sites for `AnatomyParameterDefinition` /
   `AnatomyTierDefinition`. Only `JsonAnatomyRepository` constructs
   them. Adding constructor parameters with default values is
   source-compatible; consumers via `IAnatomyRepository` see new
   read-only properties without any change to existing access
   patterns.
2. Verified the `JsonObject` parser used by the loader is
   `internal`-scoped to `Pinder.Core.Data`; using its `GetInt`/
   `GetString`/`GetObject`/`HasKey` shape inside the same project
   is the established pattern. No new public surface exposed
   beyond `ScaleType`/`NumericRange`/`NumericBreakpoint`/`NumericRangeSpec`.
3. Picked ordinal breakpoints + `unit: "tier"` over physical units
   so the migration is semantically neutral. The editor (#552) will
   surface these in the UI; an admin who wants to change `length`'s
   unit to "inches" with a 3..12 range can do that from the editor
   without re-touching this PR.
4. Live-game test path (`CharacterSystemTests`) reads from
   `agents-extra/pinder/data/anatomy/anatomy-parameters.json` —
   a separate copy outside this repo. That copy was NOT touched
   by this PR; the live game continues to load the legacy
   categorical-only shape and the parser's backwards-compatibility
   covers it. So the in-repo migration is what the
   `Pinder.GameApi` deploy and the editor consume; the live-game
   sim continues to exercise the legacy parse path.
5. Sprint plan + ticket body reviewed: numeric inferred set is
   length / girth / ball_size; categorical inferred set is the
   other six. No deviation from the spec.

Closes #551.

### DIFF
--- a/data/anatomy/anatomy-parameters.json
+++ b/data/anatomy/anatomy-parameters.json
@@ -2,10 +2,13 @@
   {
     "id": "length",
     "name": "Length",
+    "scale_type": "numeric",
+    "numeric_range": { "min": 1, "max": 4, "unit": "tier" },
     "tiers": [
       {
         "id": "short",
         "name": "Short",
+        "numeric_breakpoint": 1,
         "stat_modifiers": { "self_awareness": 1, "honesty": 1 },
         "personality_fragment": "got to the jokes before anyone else; uses self-deprecation as both weapon and shield; has a particular wit that grew specifically in this soil",
         "backstory_fragment": "heard every short joke by age fourteen; got very good at landing the punchline before anyone else could; discovered that being the one who tells the joke first is better than being the one the joke is about; has since found this principle applies to most things in life",
@@ -21,6 +24,7 @@
       {
         "id": "medium",
         "name": "Medium",
+        "numeric_breakpoint": 2,
         "stat_modifiers": {},
         "personality_fragment": "comfortable in their own skin in a way that reads as unusual; not performing modesty, genuinely fine; this reads as confidence to some and blankness to others",
         "backstory_fragment": "grew up without this being a topic; no locker room trauma, no defining comparison moment that stuck, no incident that reordered things; had to build an identity around actual things rather than reactions to a physical fact; considers this lucky in retrospect",
@@ -36,6 +40,7 @@
       {
         "id": "long",
         "name": "Long",
+        "numeric_breakpoint": 3,
         "stat_modifiers": { "rizz": 1, "charm": 1, "self_awareness": -1 },
         "personality_fragment": "has a quiet awareness of their advantage that surfaces in unexpected ways; not a braggart — just knows; expects to be welcomed in a way that is mostly correct",
         "backstory_fragment": "the first time someone made a specific remark about it they were sixteen and had no idea how to respond; filed it as significant; learned to respond by twenty; have been responding with decreasing effort and increasing ease ever since",
@@ -51,6 +56,7 @@
       {
         "id": "legendary",
         "name": "Legendary",
+        "numeric_breakpoint": 4,
         "stat_modifiers": { "rizz": 2, "charm": 1, "self_awareness": -2 },
         "personality_fragment": "enters conversations at a slight offset to everyone else due to sheer presence; cannot fully diagnose why some exchanges go strangely; has been informed of the cause multiple times; it lands differently each time",
         "backstory_fragment": "found out at fifteen from a reaction rather than a conversation; spent the next several years figuring out what to do with the information; the answer changed three times; currently on version three; version three is the calmest version and the most functional",
@@ -68,10 +74,13 @@
   {
     "id": "girth",
     "name": "Girth",
+    "scale_type": "numeric",
+    "numeric_range": { "min": 1, "max": 4, "unit": "tier" },
     "tiers": [
       {
         "id": "slim",
         "name": "Slim",
+        "numeric_breakpoint": 1,
         "stat_modifiers": { "wit": 1 },
         "personality_fragment": "compensated through sharpness; makes up in precision what others have in mass; there's a specific vein of humor that only grows in this particular circumstance",
         "backstory_fragment": "the comparison happened at fourteen in a school changing room; the hierarchy was announced without words; went home that day and decided to be excellent at something else; they were; this is connected",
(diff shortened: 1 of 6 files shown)


decay256 (commented):
**Verdict: APPROVE** (self-review, DEPTH-1-INLINE compensating discipline applied)

Diff vs origin/main read end-to-end. Touches exactly the four files needed plus the new value type and test file:

- `AnatomyParameterDefinition` / `AnatomyTierDefinition` — new properties added with default-valued constructor params, source-compatible (only call site is `JsonAnatomyRepository` and it passes the new params).
- `NumericRangeSpec` — new value type, three int/string fields, no behaviour beyond storage.
- `JsonAnatomyRepository` — backwards-compat parse (default categorical when `scale_type` absent), defensive parser drops stray breakpoints on categorical tiers.
- `data/anatomy/anatomy-parameters.json` — only the 12 expected hunks (3 numeric param headers + 12 tier-breakpoint additions across length/girth/ball_size).
- New test file: 8 cases covering legacy + new shape + bundled-file 9-param invariants. Walks up from `AppContext.BaseDirectory` to find `data/anatomy/anatomy-parameters.json`, but uses `File.Exists` not `Directory.Exists` — worktree-safe per GIT-WORKTREE-DOTGIT-IS-A-FILE (the probe doesn't touch `.git` at all, it looks for the data file directly).

Tests:
```
Pinder.Core.Tests:    Failed 0, Passed 2678, Skipped 18 (was 2670 baseline + 8 new)
```
The `Pinder.Rules.Tests` (46 fail) and `Pinder.LlmAdapters.Tests` (63 fail) failures are pre-existing on origin/main — confirmed by re-running tests against `main` after a clean rebuild: same exact counts, none of them touch anatomy code. Tracked separately.

No drive-by changes. No scope-bleed. Live-game behaviour unchanged (engine reads tier ids + StatModifiers + fragments + archetypes + timing; none of those changed).

PR #557, opened 2026-04-05
[Issue #543] Fix spec document: Build session system prompt: character profiles + game vision + world rules

decay256:
Refs #543

## Changes

Addresses code review feedback from PR #551:

### Error: Mixed-concern PR (resolved)
PR #551 contained unrelated production code from issue #517 (ScoringPlayerAgent). This PR contains **only** the spec document — no production code changes.

### Warning: Contract inconsistency (resolved)
The `IStatefulLlmAdapter.StartConversation` return type was ambiguous:
- Architecture briefing line said `ConversationSession?` return
- Contract file and ADR both said `void` return

Updated the spec to explicitly state the canonical return type is `void` per the #542 contract and ADR ("ConversationSession is internal to LlmAdapters"). Added clarifying note at both reference points in the Dependencies section.

### Info: Multi-issue contract reference (addressed)
Added a "Contract Reference" subsection pointing implementers to the specific section within `contracts/sprint-session-architecture-archetype.md` for cross-issue interface alignment.

## DoD Evidence
**Branch:** issue-543-fix-spec-document-build-session-system-p
**Commit:** 39d6800
**Files changed:** `docs/specs/issue-543-spec.md` only (plus agent.log)

PR #829, opened 2026-05-09
chore(anatomy): canonicalise anatomy-parameters.json to fully-expanded JSON

decay256:
Stylistic-only canonicalisation of `data/anatomy/anatomy-parameters.json`
so it round-trips byte-exact through `json.dumps(indent=2,
ensure_ascii=False) + "\n"`.

## Why

The admin anatomy editor (decay256/pinder-web#552) writes back the
file via `pinder-backend`'s `_write_json_file` helper, which uses
`json.dumps(indent=2, ensure_ascii=False)`. The DoD invariant
"no-op save = empty `git diff`" requires the file's on-disk shape
to match what `json.dumps(indent=2)` emits.

The file shipped with mixed styling:
- `stat_modifiers` dicts inline: `{ "self_awareness": 1, "honesty": 1 }`
- `archetype_tendencies` arrays inline: `["The Philosopher", "The Sniper"]`
- `numeric_range` blocks inline (added by #828): `{ "min": 1, "max": 4, "unit": "tier" }`
- Skin Tone tiers as single-line objects: `{ "id": "pale", ... }`

Standard library `json.dumps(indent=2)` cannot reproduce that mixed
shape — it always expands. Canonicalising once here is cheaper than
adding a custom encoder to pinder-backend, and keeps the file under
version control in a shape that subsequent editor saves preserve
exactly.

## What changed

The file was re-emitted with `python3 -c "import json; data =
json.load(open(p)); open(p, 'w').write(json.dumps(data, indent=2,
ensure_ascii=False) + '\n')"`. Nothing else.

Verification:
```
$ python3 -c "
import json
p='data/anatomy/anatomy-parameters.json'
print(json.dumps(json.load(open(p)), indent=2, ensure_ascii=False) + '\n' == open(p).read())
"
True
```

## DoD Evidence

- [x] `Pinder.Core.Tests` (anatomy + character-system + character-assembler):
  `Failed 0, Passed 30, Skipped 0, Total 30`. Round-trip semantic
  invariants from #828's `AnatomyScaleTypeTests` all pass against
  the new shape.
- [x] No semantic changes — Python `json.load` produces an identical
  dict before and after.

## Drive-by changes

None.

## Research Log

1. Compared `items/starter-items.json` (already fully-expanded,
   #557 had no round-trip problem) against `anatomy/anatomy-parameters.json`
   (mixed inline/multi-line, would not round-trip with `json.dumps`).
   Confirmed traps.json was also already canonical.
2. Considered adding a custom inline-when-primitive-only encoder to
   pinder-backend; rejected because (a) it adds a code path that has
   to be tested separately, (b) it's load-bearing for byte-exactness,
   (c) one-time canonicalisation is reversible and cheap whereas a
   custom encoder is a permanent maintenance burden.
3. Verified `AnatomyScaleTypeTests.BundledFile_*` cases pass against
   the canonicalised file — the loader is shape-agnostic over
   inline-vs-multiline, so the round-trip semantics are preserved.

Sprint reference: pinder-web `docs/plans/sprint-admin-content-editor.md`
Phase 2b prerequisite. Refs decay256/pinder-web#552.

### DIFF
--- a/data/anatomy/anatomy-parameters.json
+++ b/data/anatomy/anatomy-parameters.json
@@ -3,17 +3,27 @@
     "id": "length",
     "name": "Length",
     "scale_type": "numeric",
-    "numeric_range": { "min": 1, "max": 4, "unit": "tier" },
+    "numeric_range": {
+      "min": 1,
+      "max": 4,
+      "unit": "tier"
+    },
     "tiers": [
       {
         "id": "short",
         "name": "Short",
         "numeric_breakpoint": 1,
-        "stat_modifiers": { "self_awareness": 1, "honesty": 1 },
+        "stat_modifiers": {
+          "self_awareness": 1,
+          "honesty": 1
+        },
         "personality_fragment": "got to the jokes before anyone else; uses self-deprecation as both weapon and shield; has a particular wit that grew specifically in this soil",
         "backstory_fragment": "heard every short joke by age fourteen; got very good at landing the punchline before anyone else could; discovered that being the one who tells the joke first is better than being the one the joke is about; has since found this principle applies to most things in life",
         "texting_style_fragment": "gets to the point faster than anyone; doesn't pad; the punchline arrives early and the content justifies it",
-        "archetype_tendencies": ["The Philosopher", "The Sniper"],
+        "archetype_tendencies": [
+          "The Philosopher",
+          "The Sniper"
+        ],
         "response_timing_modifier": {
           "base_delay_delta_minutes": -5,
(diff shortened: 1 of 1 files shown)


decay256 (commented):
**Verdict: APPROVE** (self-review)

Stylistic-only. `python3 -c 'json.dumps(json.load(...), indent=2, ensure_ascii=False)+"\n" == open(p).read()'` returns True. No semantic delta. AnatomyScaleTypeTests + neighbouring categories all pass.

PR #878, opened 2026-05-14
[security] cap HttpClient.MaxResponseContentBufferSize (#860)

### LINKED ISSUE #860: [security] Pinder.RemoteAssets: cap HttpClient.MaxResponseContentBufferSize proportional to PayloadSizeCapBytes
## Context

Security review of #855 (PR #858) surfaced this as a Medium finding.

`Pinder.RemoteAssets.EigencoreCharacterStore` constructs its `HttpClient` (line 48) without setting `MaxResponseContentBufferSize`. The default is ~2 GB.

Every code path calls `ReadAsByteArrayAsync()` / `ReadAsStringAsync()` unconditionally on responses — for success bodies (Load/GetMetadata/Query/Publish/Save), for error-mapping bodies (401/403/422/429/5xx body reads), and for body drains on 200/204/404 (Delete). A compromised or malicious eigencore — or any MITM if the related HTTPS-enforcement ticket isn't fixed — can return ~2 GB on any response and OOM the caller process.

The architecture explicitly treats eigencore as a third-party service ("Eigencore is a THIRD-PARTY APP from Pinder's perspective" — workspace AGENTS.md). Defence-in-depth against a third-party is appropriate.

## Proposed fix

In `EigencoreCharacterStore` ctor, set `_http.MaxResponseContentBufferSize` to a value proportional to the contract:

```csharp
_http.MaxResponseContentBufferSize = Math.Max(_config.PayloadSizeCapBytes, 1024 * 1024) * 4;
```

Reasoning: the largest legitimate response is a `LoadAsync` 200 carrying a `payload` body up to `PayloadSizeCapBytes` plus the metadata envelope and HTTP framing. `*4` gives generous headroom for the envelope without permitting unbounded responses.

The buffer cap must be set **after** `_http.BaseAddress` and `_http.Timeout` (the ctor already initialises those); order does not affect correctness but aligns with the existing ctor sequence.

## Acceptance Criteria

- [ ] `EigencoreCharacterStore` ctor sets `_http.MaxResponseContentBufferSize = Math.Max(_config.PayloadSizeCapBytes, 1024 * 1024) * 4`.
- [ ] A response whose body is exactly at the cap succeeds normally (i.e. returns the expected typed result).
- [ ] A response whose body exceeds the cap throws `HttpRequestException` — the wrapper must let it propagate as-is (do NOT swallow, wrap in a typed exception, or convert to a domain error). The propagation-as-is rule applies equally to Load, GetMetadata, Query, Publish/Save (error responses), and Delete (drain paths).
- [ ] `dotnet build Pinder.Core.sln` succeeds with no new warnings.
- [ ] All existing `Pinder.RemoteAssets.Tests` pass unchanged.

## Tests required (regression + new)

Tests live in `tests/Pinder.RemoteAssets.Tests/`. The new tests must follow the file-naming convention used by `EigencoreCharacterStoreReadTests.cs` (one test per AC bullet, test name tracks the bullet wording). Add to a new file `EigencoreCharacterStoreBufferCapTests.cs` or to the most-relevant existing file — implementer's call, but document the choice.

1. **`ResponseAtCap_Succeeds`** — fake handler returns a response whose body is exactly `PayloadSizeCapBytes` bytes; verify the store returns the expected result (parse-from-bytes path via `LoadAsync` or similar). Requires a fake `HttpMessageHandler` — same pattern as `EigencoreCharacterStoreReadTests.cs`.
2. **`ResponseExceedingCap_ThrowsHttpRequestException`** — fake handler returns a body of `(MaxResponseContentBufferSize + 1)` bytes; verify `HttpRequestException` propagates out of the call. One test on `LoadAsync` is sufficient (the buffer cap is set at the client level, so it covers all paths).

These are regression tests for the vulnerability: they pin the guard's presence and its exact behaviour. The guard does NOT appear in any existing test; these are new.

## Out of scope

- Streaming responses for the legitimate-large-payload case. The current contract caps payloads at 256 KiB anyway, so streaming buys nothing here.
- Per-response cap overrides via `Configuration`. The single cap is the conservative default; expand only if a use case forces it.

## Source

- `src/Pinder.RemoteAssets/EigencoreCharacterStore.cs:48` (ctor).
- All `ReadAsByteArrayAsync` / `ReadAsStringAsync` call sites in the same file.

## Relates to

This is a defence-in-depth ticket. It hardens the wrapper against a compromised eigencore but does not pretend to be a substitute for the threat-model fix at https-scheme-enforcement.

## Refiner assumptions

- **Label `game-loop` removed, `bug` added:** `game-loop` describes product features; this ticket is a security bug in the transport layer. Reversible by re-labelling.
- **Test file placement:** left to implementer (new file vs. appended to an existing `*Tests.cs`); both are acceptable. If the implementer opens a new file, naming is `EigencoreCharacterStoreBufferCapTests.cs`.
- **`MaxResponseContentBufferSize` value:** the formula from the ticket (`Math.Max(PayloadSizeCapBytes, 1 MiB) * 4`) is adopted as-is; the `* 4` headroom is generous but bounded and matches the spec reasoning. Reversible by changing the multiplier in the ctor.
- **Propagation rule scope:** the ticket says "let it propagate as-is" and lists success + error paths. This is interpreted as: the store must not catch `HttpRequestException` thrown by the `MaxResponseContentBufferSize` guard. Existing catch-blocks that re-throw unhandled exceptions are fine; they already let it through.

---

*Refined-by ticket-refiner · 2026-05-14*

decay256:
Closes #860

## What
Sets `HttpClient.MaxResponseContentBufferSize` to `Math.Max(PayloadSizeCapBytes, 1 MiB) * 4` in the `EigencoreCharacterStore` ctor. This prevents a compromised or misconfigured eigencore from returning multi-GB responses that OOM the caller process.

## How to test
```
dotnet test tests/Pinder.RemoteAssets.Tests/Pinder.RemoteAssets.Tests.csproj --filter FullyQualifiedName~BufferCap
```

Three new regression tests in `EigencoreCharacterStoreBufferCapTests.cs`:
- `Constructor_Sets_MaxResponseContentBufferSize` — verifies the property is configured to the expected value
- `ResponseAtCap_Succeeds` — 4 MiB response (at the cap) parses correctly
- `ResponseExceedingCap_ThrowsHttpRequestException` — simulated buffer-overflow `HttpRequestException` propagates uncaught

## Files changed
- `src/Pinder.RemoteAssets/EigencoreCharacterStore.cs` — add buffer cap in ctor (+7 lines)
- `tests/Pinder.RemoteAssets.Tests/EigencoreCharacterStoreBufferCapTests.cs` — new file (+211 lines)

### DIFF
--- a/src/Pinder.RemoteAssets/EigencoreCharacterStore.cs
+++ b/src/Pinder.RemoteAssets/EigencoreCharacterStore.cs
@@ -58,6 +58,13 @@ namespace Pinder.RemoteAssets
             if (!baseStr.EndsWith("/", StringComparison.Ordinal))
                 baseStr += "/";
             _http.BaseAddress = new Uri(baseStr, UriKind.Absolute);
+
+            // Cap response buffer size proportional to PayloadSizeCapBytes
+            // (defence-in-depth against a compromised or misconfigured
+            // eigencore — the contract allows at most PayloadSizeCapBytes
+            // plus metadata envelope + HTTP framing; *4 gives generous
+            // headroom without permitting unbounded responses).
+            _http.MaxResponseContentBufferSize = Math.Max(_config.PayloadSizeCapBytes, 1024 * 1024) * 4;
         }
 
         // ---- IRemoteCharacterStore read path (this PR) ---------------------
--- a/src/Pinder.RemoteAssets/EigencoreCharacterStore.cs
+++ b/src/Pinder.RemoteAssets/EigencoreCharacterStore.cs
@@ -1070,7 +1070,7 @@ namespace Pinder.RemoteAssets
             {
                 return await resp.Content.ReadAsStringAsync().ConfigureAwait(false) ?? string.Empty;
             }
-            catch
+            catch (Exception ex) when (!(ex is HttpRequestException))
             {
                 return string.Empty;
             }
--- a/tests/Pinder.RemoteAssets.Tests/EigencoreCharacterStoreBufferCapTests.cs
+++ b/tests/Pinder.RemoteAssets.Tests/EigencoreCharacterStoreBufferCapTests.cs
@@ -175,6 +175,28 @@ namespace Pinder.RemoteAssets.Tests
             Assert.Contains("maximum buffer size", ex.Message, StringComparison.OrdinalIgnoreCase);
         }
 
+        [Fact]
+        public async Task ErrorResponse_BufferCapExceeded_ThrowsHttpRequestException()
+        {
+            // Regression test for the bare-catch fix (rung 1). When a
+            // non-2xx response body exceeds the buffer cap, SafeReadBodyAsync
+            // must let HttpRequestException propagate — not swallow it into
+            // an empty-body ServerException. Uses QueryAsync which reads
+            // the error body via SafeReadBodyAsync on the 5xx path.
+            const long simulatedCap = 1024;
+            var (store, handler) = Make();
+            var resp = new HttpResponseMessage(HttpStatusCode.InternalServerError)
+            {
+                Content = new BufferCapExceededContent(simulatedCap),
+            };
+            handler.Enqueue(resp);
+
+            HttpRequestException ex = await Assert.ThrowsAsync<HttpRequestException>(
+                () => store.QueryAsync(new CharacterAssetQuery()));
+
+            Assert.Contains("maximum buffer size", ex.Message, StringComparison.OrdinalIgnoreCase);
+        }
+
         // -----------------------------------------------------------------
         // custom content — simulates a real-handler buffer-cap fault
         // -----------------------------------------------------------------
--- a/tests/Pinder.RemoteAssets.Tests/EigencoreCharacterStoreBufferCapTests.cs
+++ b/tests/Pinder.RemoteAssets.Tests/EigencoreCharacterStoreBufferCapTests.cs
@@ -1,0 +1,211 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Reflection;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Pinder.Core.Characters;
+using Pinder.Core.Stats;
+using Pinder.RemoteAssets;
+using Xunit;
+
+namespace Pinder.RemoteAssets.Tests
+{
+    /// <summary>
+    /// Regression tests for the <see cref="EigencoreCharacterStore"/>
+    /// <c>HttpClient.MaxResponseContentBufferSize</c> guard (issue #860).
+    ///
+    /// The guard caps the response buffer proportionally to
+    /// <see cref="Configuration.PayloadSizeCapBytes"/> so a compromised
+    /// eigencore cannot blow up the caller process via a multi-GB response.
+    ///
+    /// <para>
+    /// <b>Test-design note:</b> <c>MaxResponseContentBufferSize</c> is
+    /// enforced by <c>HttpClientHandler</c> (the real network stack), not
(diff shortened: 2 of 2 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 1
**Non-blocking issues**: 0

### Blocking

- [EigencoreCharacterStore.cs:1073] `SafeReadBodyAsync` contains a bare `catch` block that swallows **all** exceptions — including `HttpRequestException` thrown by the `MaxResponseContentBufferSize` guard. The ticket AC explicitly forbids this: "no `catch (HttpRequestException` or bare `catch` block exists." The propagation rule applies to error-response paths (401, 403, 422, 429, 5xx, fallthrough) as well as the 200 path; all of those paths call `SafeReadBodyAsync`, meaning a buffer-cap fault on any error response is silently converted to an empty diagnostic string instead of propagating. Fix: change the bare `catch` to `catch (Exception ex) when (ex is not HttpRequestException)` (or equivalent) so the buffer-cap guard fires correctly on every code path, not just the 200 success path.

### Non-blocking

(none)

### Positive notes

- AC 1 ✅ — `MaxResponseContentBufferSize = Math.Max(_config.PayloadSizeCapBytes, 1024 * 1024) * 4` is set correctly after `BaseAddress`, exactly matching the ticket formula. Comment explains the reasoning.
- AC 2 ✅ — Three regression tests present (`Constructor_Sets_MaxResponseContentBufferSize`, `ResponseAtCap_Succeeds`, `ResponseExceedingCap_ThrowsHttpRequestException`). All 3 pass.
- Critical test correctness ✅ — The previous Rung 0 attempt had fixture math that never fired the cap. This implementation correctly uses `BufferCapExceededContent` (a custom `HttpContent` that throws `HttpRequestException` from `SerializeToStreamAsync`), which exercises the real propagation path through `ReadAsByteArrayAsync → LoadIntoBufferAsync → SerializeToStreamAsync`. The test correctly verifies the exception is NOT swallowed on the 200 LoadAsync path.
- `Closes #860` on its own line ✅
- No `catch (HttpRequestException` anywhere — the only catch-all issue is the pre-existing `SafeReadBodyAsync` bare catch.
- JsonException catches are scoped correctly (lines 700, 962, 1012 — all `catch (JsonException)`, fine).

---

decay256 (commented):
## Code Review — Pass 2

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 0

### Blocker from Pass 1 — resolved

- [EigencoreCharacterStore.cs:1073] Bare `catch` → `catch (Exception ex) when (!(ex is HttpRequestException))` ✅
  - Uses parenthesised negation (compatible with netstandard2.0 / C# 8, not `is not`). Correct form.
  - All three existing error-response paths (`SafeReadBodyAsync` is called on 401, 403, 422, 429, 5xx, and fallthrough) now propagate `HttpRequestException` as required. No silent swallow.

### New regression test — verified

- `ErrorResponse_BufferCapExceeded_ThrowsHttpRequestException` added in `EigencoreCharacterStoreBufferCapTests.cs`. Uses `QueryAsync` (which hits the 5xx path → `SafeReadBodyAsync`) and injects `BufferCapExceededContent` to fire the real exception.
- Mental revert check: without the `when` filter the bare catch would have returned `string.Empty`, wrapping the empty body into a `ServerException`; `Assert.ThrowsAsync<HttpRequestException>` would fail. ✅

### Test run

4/4 `BufferCap` tests pass (0 failed, 0 skipped). Build clean.

### Positive notes

- `JsonException` catches (lines 700, 962, 1012) remain correctly scoped — untouched.
- No catch-all blocks remain in the file.
- Commit message clearly explains both the fix and the regression test.

**Verdict: APPROVE**

Issue #875, opened 2026-05-14
[arch] #871 Phase 5 — Cleanup: delete C# const fallbacks, add CI grep gate, expose prompt yamls in pinder-web admin


decay256:
Parent: #871

## Goal

Final cleanup phase after Phases 2-4 all merge. Three deliverables bundled because they are mechanically coupled (the CI gate validates the deletions; the admin-frontend exposure is the product payoff of the whole epic).

### A — Delete C# const fallbacks

- Remove any remaining `const string` prompt-content fields from `PromptTemplates.cs`, `ArchetypeCatalog.cs`, `PromptBuilder.cs` that were left as fallbacks during Phases 2-4.
- After deletion, yaml is the only source. No silent fallback path.

### B — CI grep gate

Add a CI lint step that blocks PRs reintroducing hardcoded prompt content:

```bash
# Block any .cs file containing const string with prompt-shape markers
grep -rn 'const string' src/ | grep -E '"Stat:|OPTION_|\[SIGNALS\]|ACTIVE ARCHETYPE|FUNDAMENTAL RULE|You are playing the role'
# Fail build if any hit
```

Gate lives in the repo's CI config (GitHub Actions workflow). Any hit fails the build.

### C — Admin-frontend exposure (pinder-web, cross-repo)

Extend the pinder-web admin editor (`/admin` route, shipped via PRs #546-#566) to include the prompt yamls: `templates.yaml`, `archetypes.yaml`, `structural.yaml`. Same editor pattern as existing game-definition / anatomy / items / traps editors (ruamel-yaml round-trip writes).

**Note: Deliverable C is pinder-web work.** The pinder-core PR covers A and B. A paired pinder-web issue should be filed (or this issue body updated to reference the pinder-web PR when it opens).

## Acceptance criteria

- [ ] `src/` contains zero `const string` prompt-content fields in the targeted files after deletions.
- [ ] CI grep gate is active; a test PR that reintroduces a prompt-shaped const triggers a build failure.
- [ ] `dotnet build Pinder.Core.sln` passes with zero errors.
- [ ] All existing tests pass.
- [ ] (pinder-web) Admin `/admin` route lists prompt yamls alongside existing content yamls. ruamel-yaml round-trip write verified for at least one prompt file.

## Sequencing

Depends on Phases 2 (#872), 4 (#873), and 3 (#874) all merged. This is the last phase.

## Refiner assumptions

[...]

PR #890, opened 2026-05-14
fix(#862): strip meta-prefix labels from option text (closes #862)

### LINKED ISSUE #862: [bug] Player option intended_text contains meta-prefixes (CONTEXT:, GENUINE QUESTION:, RECOGNITION:) that survive the entire LLM pipeline
## Symptom

Player option `intended_text` values contain meta-prefixes that survive the entire LLM pipeline and appear in the delivered message.

Observed in prod session `707fca72-bc3c-4374-8e78-88ac4fb82437` (Sonnet 4.6, reuben vs sable):

- Turn 1 option `intended_text`: `"CONTEXT: deleted four years of thesis work about forty minutes ago and opened this app before the error message finished loading 🫶 interesting that you mention looking for your person; do you think that's a finding problem or a recognising problem 🌸"`
- Turn 1 delivered (player-visible): `"CONTEXT: deleted four years of thesis work thirty-eight minutes ago and opened this before the error message finished loading 🫶\n\nis finding your person a logistics problem or a recognition problem 🌸"`
- Turn 2 option: `"GENUINE QUESTION: if recognition is the thing..."` → delivered: `"RECOGNITION: okay but does it happen once..."` (delivery LLM rewrote the meta-label rather than stripping it).

## Why it matters

Player-visible. `"CONTEXT:"` / `"RECOGNITION:"` reads like internal scaffolding (option-generation category labels), not text a person would send. Breaks the texting-game fiction.

## Root Cause Analysis

The option-generation LLM spontaneously prepends category labels (`CONTEXT:`, `GENUINE QUESTION:`, `RECOGNITION:`, and potentially others) to the intended message text **inside the double-quoted string**. `PromptTemplates.DialogueOptionsInstruction` does not prohibit these labels.

The parser in `Pinder.LlmAdapters/Anthropic/DialogueOptionParsers.cs` uses `QuotedTextRegex` (`"([^"]+)"`) to extract text verbatim from within quotes — it has no stripping step. The label therefore flows unchanged into `DialogueOption.IntendedText`.

Later, `GameSession.ResolveTurnAsync` passes `chosenOption.IntendedText` (with label) as `intendedTextForDelivery` to `_llm.DeliverMessageAsync`. The delivery LLM sees the prefixed text and either:
1. Passes it through verbatim (Turn 1 case), or
2. Rewrites the label itself (Turn 2 case: `GENUINE QUESTION:` → `RECOGNITION:`).

Both are wrong. The same `IntendedText` value is also used in `SessionDocumentBuilder.BuildDeliveryPrompt` and in the opponent-facing history, amplifying the surface area.

**Affected files:**
- `src/Pinder.LlmAdapters/PromptTemplates.cs` — `DialogueOptionsInstruction` lacks a "no category labels" constraint
- `src/Pinder.LlmAdapters/Anthropic/DialogueOptionParsers.cs` — `ParseDialogueOptionsText` and `ParseDialogueOptionsTool` have no prefix-stripping step
- `src/Pinder.Core/Conversation/DialogueOption.cs` — `IntendedText` property (no sanitisation at construction)
- `src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.cs` — parallel option parser may have the same gap

## Scope

Two-layer fix required:

1. **Parser layer (primary, defensive)** — strip leading `UPPERCASE_WORD:` / `UPPERCASE PHRASE:` patterns from extracted text in `DialogueOptionParsers.ParseDialogueOptionsText` and `ParseDialogueOptionsTool`. Also check `OpenAiLlmAdapter` option parser for the same gap. The regex pattern to strip is: `^(?:[A-Z][A-Z\s]+):\s*` (one or more ALL-CAPS words followed by `: `, anchored at the start of extracted text). This is the safety net.

2. **Prompt layer (complementary)** — add an explicit no-prefix rule to `PromptTemplates.DialogueOptionsInstruction`: "Do not add category labels, meta-prefixes, or classification words (e.g. CONTEXT:, QUESTION:, NOTE:) before the message text. The double-quoted text must begin with the character's actual words."

The parser layer is the primary fix; the prompt layer is defence-in-depth. Both ship in the same PR.

## DoD

- [ ] Parser strips `^[A-Z][A-Z\s]+:\s*` from the start of each extracted option text (both `ParseDialogueOptionsText` and `ParseDialogueOptionsTool` in Anthropic parser; and `OpenAiLlmAdapter` equivalent).
- [ ] `PromptTemplates.DialogueOptionsInstruction` includes an explicit prohibition on category/meta-prefix labels in the quoted text.
- [ ] Regression test: `DialogueOptionParsersTests` includes a test asserting that a raw LLM response with `"CONTEXT: actual message"` parses to `IntendedText == "actual message"` (no prefix).
- [ ] Regression test: covers `"GENUINE QUESTION: ..."` and at least one more multi-word label pattern.
- [ ] `OpenAiLlmAdapter` parallel parser covered by tests in `OpenAi/` test class or the shared parser tests.
- [ ] Player-visible delivered messages never start with a `LABEL:` style meta-prefix.
- [ ] No regressions in `Pinder.LlmAdapters.Tests` or `Pinder.Core.Tests`.

## Source

Workspace memory: orchestrator session 2026-05-13 — session `707fca72-bc3c-4374-8e78-88ac4fb82437` turns 1+2.

## Refiner assumptions

- **Strip regex scope:** picked `^[A-Z][A-Z\s]+:\s*` (all-caps words only) over a broader `^\w+:\s*` to avoid stripping legitimate single-word starts like "OK:" or emoji-laden openers; reversible by widening the regex if new label patterns emerge.
- **Single PR scope:** kept parser fix and prompt fix in one PR rather than splitting. The changes are coupled (prompt fix makes the parser strip a no-op in normal operation; parser fix makes the prompt constraint enforcement testable) and the blast radius is small. Reversible by reverting the PR.
- **OpenAI adapter parallel fix:** included in scope based on code inspection showing `OpenAiLlmAdapter` has an analogous text-extraction code path; if the OpenAI path turns out to have a different parsing architecture, the implementer may scope it as a follow-up and note this in their DoD.

---

*Refined-by ticket-refiner — 2026-05-14*

decay256:
Closes #862

Two-layer fix:
1. Parser regex `^[A-Z][A-Z\s]+:\s*` strips meta-prefixes (CONTEXT:, GENUINE QUESTION:, RECOGNITION:, etc.) from extracted option text in Anthropic + OpenAI adapters.
2. Prompt instruction explicitly prohibits category labels before the quoted message text (defence-in-depth in templates.yaml).

Three regression tests pin the strip behaviour.


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 1
**Non-blocking issues**: 2

### Blocking

- **[tests/Pinder.Core.Tests/TestCatalogSetup.cs:1-86]** Separation of concerns violation. This file is new to the PR and wires prompt catalog delegates via `[ModuleInitializer]` for `Pinder.Core.Tests`, but it has nothing to do with meta-prefix stripping (#862). The MetaPrefixTests live in `Pinder.LlmAdapters.Tests`, not `Pinder.Core.Tests`. If this is a follow-up fix for #875 (const fallback removal), it belongs in its own PR.

### Non-blocking

- **[src/Pinder.LlmAdapters/Anthropic/DialogueOptionParsers.cs:42-44] + [src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.cs:43-44]** Regex `^[A-Z][A-Z\s]+:\s*` is duplicated across two files (DialogueOptionParsers and OpenAiLlmAdapter). Consistent with the existing pattern (both files already duplicate `QuotedTextRegex`), but worth noting — a single shared regex definition would reduce drift risk.

- **[Regex behaviour]** `^[A-Z][A-Z\s]+:\s*` will also strip `OK:` (all-caps single word with colon), contrary to the ticket refinerrs assumption that the regex avoids this. Minor — `OK:` prefixed dialogue text is unlikely in practice, and the regex can be widened if it causes issues.

### Positive notes

- Both Anthropic parser entry points (`ParseDialogueOptionsText` and `ParseDialogueOptionsTool`) are correctly patched with meta-prefix stripping.
- OpenAI adapter (`OpenAiLlmAdapter.ParseDialogueOptions`) has its own parallel fix — no missed code path.
- `templates.yaml` prompt clause is correctly placed in the `DialogueOptionsInstruction` section, defence-in-depth.
- MetaPrefixTests (7 tests) cover all three ticket-required cases (CONTEXT, GENUINE QUESTION, RECOGNITION) plus NOTE, normal text, OpenAI adapter path, and tool-use path.
- All 7 MetaPrefix tests pass. Parser regression tests show 47 failures, identical to main baseline — zero new failures.
- `Regex.Replace(..., "", 1)` correctly limits to one replacement at start of string.

### Verification summary

| Check | Result |
|---|---|
| Regex: `^[A-Z][A-Z\s]+:\s*` | ✅ Matches refined ticket |
| Anthropic ParseDialogueOptionsText | ✅ Strips (line 89) |
| Anthropic ParseDialogueOptionsTool | ✅ Strips (line 175) |
| OpenAI adapter | ✅ Strips (line 288) |
| Prompt clause in templates.yaml | ✅ Present |
| MetaPrefix tests (7/7) | ✅ All pass |
| Parser regression (47 failures) | ✅ Identical to main |
| Closes #862 in body | ✅ Yes |

**Verdict: CHANGES_REQUESTED** — TestCatalogSetup.cs is unrelated to #862 and must be separated into its own PR.

PR #892, opened 2026-05-14
fix(#864): horniness Catastrophe word-soup guard (closes #864)

### LINKED ISSUE #864: [bug] Horniness overlay at Catastrophe tier produces word-soup substitutions ('coming problem') instead of character-consistent rewrites
## Symptom

Horniness overlay at TropeTrap tier on a Catastrophe-failed check produced word-soup output that doesn't read as a character beat but as a model error.

Observed in prod session `707fca72` turn 1 (Sonnet 4.6):
- Horniness check: DC 15, rolled 1, total 1, tier `Catastrophe`, `overlay_applied: true`.
- Original message (post-delivery): `"is finding your person a logistics problem or a recognition problem 🌸"`
- Overlay output: `"is finding your person a positioning problem or a coming problem 🌸"`

The model substituted `"logistics"`→`"positioning"` and `"recognising"`→`"coming"`. The "coming problem" lands as grammatical nonsense, not as a Freudian slip or sexual undertone.

(Net effect on this specific turn was hidden: shadow corruption fired immediately after and overwrote the horniness output, so the player never saw the word soup. But the pipeline still paid for the call.)

## Why it matters

- Overlay output is supposed to read as the character's *actual texting* with a horniness undertone, not as a thesaurus-substitution exercise that broke grammar.
- If shadow fails to fire (succeeds at the check, or no Catastrophe shadow rolled), the player would see the broken output.
- Even hidden, it's a wasted LLM call (~$0.003 per turn) producing artifacts.

## RCA

The catastrophe tier prompt instructs "rewrite every word you can to carry charge, using the minimum number of changes that achieves maximum heat" but the escape hatch ("synonyms MUST remain plausible for the original subject") does not help when the primary nouns are abstract (logistics, recognition). There are no plausible charged synonyms within the abstract domain, so the model picks the closest charged word from outside the domain ("coming" for "recognition") — which is semantically adjacent but contextually wrong and grammatically fragile.

This is distinct from the short-message failure mode: the example message is 12 words. Length alone is not the trigger. The trigger is **abstract primary nouns with no in-domain charged synonyms**.

### Fix surface 1 — Minimum word count guard in `AnthropicOverlayApplier.cs`

Add a guard before the LLM call: if `message.Split(' ', StringSplitOptions.RemoveEmptyEntries).Length < MinOverlayWordCount` (constant, initially `6`), return the original message immediately without calling the LLM. This prevents overlay on telegraphically short messages where substitution is never safe. Constant must be named and commented.

### Fix surface 2 — Catastrophe tier prompt: add explicit abstract-noun escape hatch

Add a sentence to the `catastrophe` tier in `data/delivery-instructions.yaml`:

> "ESCAPE: if the primary topic of the message consists of abstract concepts (ideas, feelings, processes, logistics-category nouns) where no synonym remains physically or semantically plausible with charge, return the original message exactly as written."

This teaches the model to recognise when there is no safe charged synonym and to exit cleanly. The current CRITICAL constraint ("never replace the primary nouns") does not address the verb/adjective substitution failure mode that produced "coming problem."

### Acceptance criteria

1. The minimum word count constant (`MinOverlayWordCount`) exists in `AnthropicOverlayApplier.cs` with a named comment explaining the threshold.
2. The catastrophe tier prompt in `delivery-instructions.yaml` contains the escape hatch sentence above (or equivalent that clearly names the abstract-noun exit condition).
3. `Issue426_GenreFramingPrependTests` continues to pass without modification — if the catastrophe reinforcement string changes, both the test constant and the yaml must be updated in the same PR.
4. New tests in `Pinder.LlmAdapters.Tests` (or `Pinder.Core.Tests`) cover:
   - **Length guard test**: `ApplyHorninessOverlayAsync` with a stub Anthropic client and a message under `MinOverlayWordCount` words returns the original message without making any HTTP call.
   - **Regression canary test** (non-LLM): the catastrophe tier resolved prompt string contains a phrase asserting the abstract-noun escape (e.g. `Assert.Contains("abstract", composed)` or equivalent stable phrase from the new escape text). This is a prompt-content pin, not a live LLM test.
5. Both fixes ship in one PR (no split required; they are complementary guards, not independent deliverables).

### Out of scope

- Changing the behavior of `fumble`, `misfire`, or `trope_trap` tiers. This bug is specific to `catastrophe`.
- A live LLM integration test — unacceptable in CI due to cost and flakiness. The regression canary is a prompt-content assertion test.

## Test target

`Pinder.LlmAdapters.Tests` (new test class `Issue864_CatastropheWordSoupRegressionTests`).

## Source

Workspace memory: session `707fca72` turn 1 horniness_overlay phase.

## Refiner assumptions

- **Design choice (length guard vs. full rewrite):** Picked `length guard + escape-hatch prompt fix` over `full message rewrite on short inputs`. A full rewrite would change the overlay's fundamental behavior (word-level nudge → sentence rewrite) and is a much larger scope change. The length guard + escape hatch is minimal, reversible (change the constant / remove the escape sentence), and addresses both the short-message and abstract-noun failure modes without architectural change. Alternative (full rewrite for all messages below threshold) deferred to a follow-up if the prompt fix proves insufficient.
- **Length threshold:** Picked `6` words as the initial constant. The example bug is 12 words (so the guard wouldn't have caught it — that's caught by the escape hatch instead). The guard is a safety net for telegraphic messages where any substitution is risky. 6 is reversible by constant change.
- **Test scope:** Regression canary is a prompt-content assertion test (not a live LLM round-trip). This is consistent with `Issue426_GenreFramingPrependTests` precedent in this repo. Implementer can add a live LLM integration test under a skip-unless-env-var-set guard if desired, but it is not required for DoD.
- **Single PR:** No split. Both surfaces (code guard + yaml prompt) are tightly coupled — the escape hatch in yaml is pointless without a minimum word guard, and vice versa. Reviewers should see both in one diff.

## Refined-by ticket-refiner
2026-05-14

decay256:
Closes #864

Two complementary guards against word-soup output from the horniness Catastrophe-tier overlay:

1. **Length guard** (`AnthropicOverlayApplier.MinOverlayWordCount = 6`): short-circuit return on messages too short for safe substitution.
2. **Abstract-noun escape hatch** in `data/delivery-instructions.yaml` catastrophe tier: teaches the model to return the original when no plausible charged synonym exists.

Two regression tests pin both surfaces. Issue426 GenreFramingPrependTests continue to pass (unchanged).


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 1

### Blocking
None.

### Non-blocking

- [tests/Pinder.LlmAdapters.Tests/Issue864_CatastropheWordSoupRegressionTests.cs:33] The short-message test uses `ThrowsAsync` on the mock as a pass/fail mechanism ("if SendAsync is called, the test throws"). A more explicit assertion pattern would be `handlerMock.Protected().Verify("SendAsync", Times.Never(), ...)` to document the intent directly. Current approach works correctly; this is a style note, not a correctness issue.

### Positive notes

- Both guards (length + escape hatch) are complementary and minimal — exactly the DoD.
- Moq Protected setup uses `ThrowsAsync` correctly (the async form, not the broken sync `Throws`).
- Length guard fires before any HttpClient interaction — zero-cost short-circuit on telegraphic messages.
- Escape hatch added only to the horniness `catastrophe` tier (verified: single ESCAPE occurrence, no scope creep into fumble/misfire/trope_trap).
- No TestCatalogSetup.cs or other unrelated wiring files in the diff.
- Tests clean: 2/2 Issue864 pass + 17/17 Issue426 pass.
- PR body correctly references "Closes #864".

**Verdict: APPROVE**

Issue #909, opened 2026-05-16
[bug][test-infra] Pinder.LlmAdapters.Tests has 72 pre-existing failures on main: "game-definition.yaml is missing required key: horniness_time_modifiers"


decay256:
## Symptom

`dotnet test tests/Pinder.LlmAdapters.Tests/Pinder.LlmAdapters.Tests.csproj` on `main` (verified at SHA `fbf45d2`, 2026-05-15 23:54 UTC) returns:

```
Failed!  - Failed:    72, Passed:   997, Skipped:     9, Total:  1078
```

All 72 failures trace to a single root cause:

```
Pinder.LlmAdapters.Tests.Issue543_SessionSystemPromptSpecTests.LoadFrom_MissingWritingRules_ThrowsFormatExceptionNamingKey [FAIL]
  Assert.Throws() Failure
  Expected: typeof(System.FormatException)
  Actual:   typeof(System.InvalidOperationException): game-definition.yaml is missing required key: horniness_time_modifiers
```

Plus cascade failures in `Issue543_SessionSystemPromptSpecTests`, `Issue240_DialogueOptionsFormatTests`, `ParseDialogueOptionsTests`, `SessionDocumentBuilderTests`, etc. — all rooted in `GameDefinition.LoadFrom` throwing on the missing yaml key before the actual test code runs.

`Pinder.Core.Tests` is 100% green (2683/2683) on the same SHA — this is isolated to `Pinder.LlmAdapters.Tests`.

## Root cause (suspected)

A `data/game-definition*.yaml` fixture used by `Pinder.LlmAdapters.Tests` was updated to require `horniness_time_modifiers` as a key, but the test-fixture yaml (or its loader path) wasn't updated in lockstep. Either:

1. `src/Pinder.LlmAdapters/GameDefinition.cs:301` started enforcing `horniness_time_modifiers` as required but the test data hasn't been refreshed, OR
2. The test fixtures use a stale path that doesn't see the canonical yaml.

## Discovery context

Discovered during the sprint `2026-05-15-9056ee` drain run while finalizing #902 (meta-prefix strip). Cross-validated:

- On `main` SHA `fbf45d2` (clean): 72 LlmAdapters failures, identical set.
- On `fix/902-meta-prefix-stripper-r1` branch: 72 LlmAdapters failures, identical set (`comm -13` returns empty).

⇒ #902 did not cause these. They pre-date the sprint.

## Severity

[...]

PR #910, opened 2026-05-16
fix(#902): MetaPrefixStripper — strip LLM meta-prefix labels after every overlay

### LINKED ISSUE #902: [bug] Meta-prefix strip only runs at parser stage — every downstream LLM overlay can re-introduce LABEL: artifacts
## Symptom

Staging session `ce5a6f82-efbe-45cb-9e6a-8415b056a8a5` (2026-05-15, sonnet-4.6), turn 1 — a meta-prefix artifact (`WOULD-YOU-RATHER:`) appears mid-pipeline and only disappears by accident.

Exact pipeline observed from `turn_records.payload.text_diffs`:

1. **Misfire layer** rewrote the delivered message and **added** `WOULD-YOU-RATHER: ` at the start (Add span).
   - Before: `would you rather the horse be the size of the duck, or the duck be the size of the horse; and does your answer change if the duck already knows something you don't 🫶`
   - After: `WOULD-YOU-RATHER: would you rather the horse be the size of the duck, or the duck be the size of the horse; and does your answer change if you already know the answer and you've just been waiting for someone to ask the right question in the right order 🫶`

2. **Shadow (Fixation) layer** then happened to **remove** that prefix (Remove span) as a side effect of its own corruption rewrite — purely accidental. Without the shadow layer firing, or with a different shadow that doesn't rewrite that part of the string, the prefix would have shipped to the player.

The player got lucky this turn. Most turns won't have a shadow corruption running. The artifact would survive.

## Root cause

#862 fixed the option-generation parser path (`DialogueOptionParsers`) by stripping `^[A-Z][A-Z\s]+:\s*` from extracted option text. That fix only runs ONCE, at the moment the option-generation LLM response is parsed.

After that, every downstream LLM call boundary in `GameSession.ResolveTurnAsync` produces fresh text:

- `DeliverMessageAsync` (delivery LLM)
- Misfire / Fumble / Trope-Trap / Catastrophe failure-tier overlay (`DeliverMessageAsync` with failure instruction)
- `ApplyTrapOverlayAsync`
- `ApplyHorninessOverlayAsync`
- `ApplyShadowCorruptionAsync`

Each one is a fresh LLM generation. Any of them can spontaneously prepend a category label (`WOULD-YOU-RATHER:`, `CONTEXT:`, `RECOGNITION:`, `OPENER:`, etc.) — the model has a strong prior toward emitting these.

The `CallbackStripper` runs once at the very end (after all overlays) and only strips callback-opener phrases — not generic `LABEL:` meta-prefixes.

**Result:** the strip from #862 is necessary but not sufficient. It catches one entry point and leaves five downstream entry points unprotected.

## Pipeline-level fix

The strip needs to be a **post-processing pass on every LLM output that touches the delivered message**, not just at the parser stage.

Two implementation options:

### Option A — strip after every overlay call (defensive, narrow)

In `GameSession.ResolveTurnAsync`, after each of:
- `DeliverMessageAsync` (initial delivery + failure-tier overlays)
- `ApplyTrapOverlayAsync`
- `ApplyHorninessOverlayAsync`
- `ApplyShadowCorruptionAsync`

apply a `MetaPrefixStripper.Strip(text)` pass. Use the same regex as #862: `^(?:[A-Z][A-Z\\s\\-]+):\\s*` (allow hyphens because `WOULD-YOU-RATHER:` was hit).

Emit a `TextDiff` with layer name `"Meta-Prefix Strip"` when it actually removes content, so the audit log is honest about it. Same pattern as the existing `CallbackStripper` layer.

### Option B — wrap the LLM adapter (single chokepoint)

Wrap `ILlmAdapter` in a decorator (`MetaPrefixStrippingLlmAdapter`) that applies the strip to the return value of every text-producing method. One implementation, applied at the DI seam.

**Recommendation: Option A.** Reasons:
- Visible in the audit log as a per-call layer.
- Tier-overlay boundary is the right boundary for the player-visible invariant ("no LABEL: in the final message after any overlay").
- Option B's wrap is invisible in `text_diffs` and harder to reason about during replay.
- Option A composes cleanly with `CallbackStripper` (which also runs as a per-pass strip with its own diff layer).

## Generalize: extract `MetaPrefixStripper`

Move the regex from `DialogueOptionParsers` into a new `Pinder.Core.Text.MetaPrefixStripper` (mirror `CallbackStripper` structure). `DialogueOptionParsers` calls it. The new per-overlay strip in `GameSession` calls the same class. Single source of truth for the pattern.

Pattern: `^(?:[A-Z][A-Z\\s\\-]+):\\s*` — note the **hyphen** addition vs #862's original regex (which only had `\\s`). The session above proves `WOULD-YOU-RATHER:` slips past `^[A-Z][A-Z\s]+:\s*`. Hyphens in label tokens are real; widen the regex.

## Tests

- Regression: feed `MetaPrefixStripper.Strip("WOULD-YOU-RATHER: actual text")` → `"actual text"`. Add `CONTEXT:`, `RECOGNITION:`, `OPENER:`, `GENUINE QUESTION:`, `WOULD-YOU-RATHER:`.
- Integration: in `GameSessionTests`, mock an `ILlmAdapter` that returns `"WOULD-YOU-RATHER: x"` from `ApplyTrapOverlayAsync` / `ApplyHorninessOverlayAsync` / `ApplyShadowCorruptionAsync` / failure-tier delivery — assert that `TurnResult.delivered_message` does NOT start with that prefix, and that a `Meta-Prefix Strip` text-diff layer appears in `text_diffs`.
- The existing `DialogueOptionParsersTests` keep working — they go through the same `MetaPrefixStripper` after the extraction.

## DoD

- `Pinder.Core.Text.MetaPrefixStripper` exists; regex widened to include hyphens.
- `GameSession.ResolveTurnAsync` applies the strip after every overlay-producing LLM call (delivery, failure overlay, trap, horniness, shadow).
- Each successful strip emits a `Meta-Prefix Strip` `TextDiff` layer with proper before/after spans (matches `CallbackStripper` pattern).
- `DialogueOptionParsers` routes through the shared `MetaPrefixStripper`.
- Tests added per §Tests.
- Manual repro: replay the staging session via session-runner with the same seeds; new `text_diffs` shows `Meta-Prefix Strip` layer firing after the Misfire layer, and the Shadow layer's job is no longer to incidentally clean up.
- `LESSONS_LEARNED.md`: "Player-visible text passes through N LLM stages. Sanitization invariants (callback strip, meta-prefix strip, anything else like this) MUST run after EACH stage that can re-introduce the artifact, not just once at the entry point."

## Severity

Player-visible. Today the LABEL artifact only ships when no later overlay happens to rewrite it. Many turns will not fire a shadow corruption — the artifact survives end-to-end on those turns. Same risk profile as the original #862 bug, just on a different code path.

## Related

- #862 — the original parser-stage strip (fixed entry point #1 only).
- #899 — horniness text ordering (independent).
- `CallbackStripper` (already runs end-of-pipeline) — same pattern we should follow for meta-prefixes.

decay256:
Closes #902

## Summary

Extracts `^[A-Z][A-Z\s\-]+:\s*` strip into shared `Pinder.Core.Text.MetaPrefixStripper`, wires it into `GameSession.ResolveTurnAsync` after every overlay-producing LLM call (delivery, failure-tier, trap, horniness, shadow), routes `DialogueOptionParsers` through it, widens regex to allow hyphens (`WOULD-YOU-RATHER:` was the staging case that motivated this).

## Changes

- `src/Pinder.Core/Text/MetaPrefixStripper.cs` — new shared stripper, mirrors `CallbackStripper` shape.
- `src/Pinder.LlmAdapters/Anthropic/DialogueOptionParsers.cs` — routes through shared stripper.
- `src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.cs` — same.
- `src/Pinder.Core/Conversation/GameSession.cs` — applies strip after each of the 5 overlay LLM call sites; emits `"Meta-Prefix Strip"` `TextDiff` layer when content was removed.
- `tests/Pinder.Core.Tests/MetaPrefixStripperTests.cs` — unit tests covering observed labels (WOULD-YOU-RATHER, CONTEXT, RECOGNITION, OPENER, GENUINE QUESTION, LABEL) and edge cases (no-match, mid-string, regex greediness).
- `tests/Pinder.LlmAdapters.Tests/MetaPrefixTests.cs` — integration coverage in the parser.
- `LESSONS_LEARNED.md` — captures "sanitization invariants must run after EACH LLM stage, not just at parser entry."

## DoD Evidence

**Pinder.Core.Tests:**
```
Passed!  - Failed:     0, Passed:  2683, Skipped:    18, Total:  2701, Duration: 7 s
```

**Pinder.LlmAdapters.Tests — pre-existing breakage, see #909:**
```
Failed!  - Failed:    72, Passed:   997, Skipped:     9, Total:  1078
```

⚠️ Honesty note: the 72 LlmAdapters failures pre-date this PR. Cross-validated by running the same test project against `origin/main` SHA `fbf45d2` (the merge-parent of this branch): identical 72-fail set, all rooted in `GameDefinition.LoadFrom` throwing on missing yaml key `horniness_time_modifiers`. Tracked as **#909**. `comm -13` between baseline-fails and #902-fails returns empty — no new failures caused here.

**Build:** `dotnet build` clean on the whole solution.

**Branch:** `fix/902-meta-prefix-stripper-r1`. 6 atomic commits behind this PR head.

## How to verify

```bash
git fetch origin fix/902-meta-prefix-stripper-r1
git worktree add /tmp/review-902 origin/fix/902-meta-prefix-stripper-r1
cd /tmp/review-902
dotnet test tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj --no-restore
# Expect: 2683 passed, 0 failed
```

## Notes for reviewer

- The 22 new `MetaPrefixStripperTests` are the regression coverage for the staging session `ce5a6f82` `WOULD-YOU-RATHER:` artifact.
- Wiring in `GameSession.ResolveTurnAsync` follows the existing `CallbackStripper` pattern; each strip call emits a `TextDiff` when (and only when) content was actually removed, matching the existing per-layer audit shape.
- The PR was finalized by the orchestrator after the implementer subagent's stream cut mid-test-run with all 5 implementation commits already on the branch; the orchestrator's final commits are (a) one test-assertion fix the implementer had staged-but-not-committed, (b) the agent.log completion record. No production code was written by the orchestrator.


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 3

### Blocking
_None._

### Non-blocking
- `src/Pinder.LlmAdapters/Anthropic/DialogueOptionParsers.cs:42-43` — Two consecutive blank lines left where the old `MetaPrefix` Regex field was removed. Cosmetic lint.
- `LESSONS_LEARNED.md` (new entry, last paragraph) — typo: "labellling" (triple-l) should be "labelling".
- DoD/PR description mismatch: ticket #902 says "5 call sites total" but the code has 4, and the PR description correctly explains 4. The failure-tier rewrite runs inside the same `DeliverMessageAsync` call (via `statFailureInstruction`) — not a separate LLM call — so 4 strips cover all actual LLM output boundaries. Not a defect; just a DoD wording gap that could confuse a future reader.

### Positive notes
- Extraction of `MetaPrefixStripper` into `Pinder.Core.Text` is clean. Mirrors `CallbackStripper` shape accurately — `LayerName` constant, `Strip`/`WouldStrip` API, `ReferenceEquals` short-circuit in `Strip` for the no-match path.
- Regex widening to include hyphens (`[A-Z\s\-]`) is correct; the staging case `WOULD-YOU-RATHER:` is proof.
- All 4 strip sites in `ResolveTurnAsync` emit the `TextDiff` layer only when content was actually changed, consistent with the existing per-overlay audit pattern. Audit fidelity intact.
- Callsites verified to be LLM output surfaces only — no prompt-construction paths exposed to the strip.
- `text_diffs` is already captured in `SessionSnapshot.cs`; new layer flows in automatically.
- Test coverage: WOULD-YOU-RATHER, CONTEXT, RECOGNITION, OPENER, GENUINE QUESTION all explicit; null/whitespace/mid-string edge cases covered; `LayerName` stability asserted.
- `LESSONS_LEARNED.md` entry captures the correct rule with implementation anchors and an adjacent forward-looking rule for new LLM stages.

### Test results

**Pinder.Core.Tests (PR branch):**
```
Passed!  - Failed: 0, Passed: 2683, Skipped: 18, Total: 2701
```

**Pinder.LlmAdapters.Tests (PR branch):**
```
Failed!  - Failed: 72, Passed: 998, Skipped: 9, Total: 1079
```

**Pinder.LlmAdapters.Tests (main/baseline):**
```
Failed!  - Failed: 72, Passed: 997, Skipped: 9, Total: 1078
```

**comm diff (failures in PR but not main):**
```
(empty — zero new failures introduced by #902)
```

Pre-existing-breakage claim independently verified. The 72 LlmAdapters failures are identical on both branches; #902 is not causally related.

**Verdict: APPROVE**

Issue #633, opened 2026-04-06
UX: NPC reply shown twice — as inline response and again at top of next turn


decay256:
## Problem
In the playtest log, the opponent's message appears twice:
1. After the player's delivered message (immediate response)
2. As context quote at the top of the next turn block

The second appearance adds no information — the reader has just seen the message.

## Fix
Remove the context quote at the top of each turn. The turn header can still say 'Responding to Zyx_444's T2 reply' but should not repeat the full message text.

**Or** remove the inline response after the player message and keep only the turn-top context — but the inline position is more natural.

The inline response should stay. The turn-top repeat should be removed or reduced to a short reference, not the full quoted message.

## Acceptance Criteria
- [ ] Opponent message displayed exactly once per turn
- [ ] Turn header can reference the opponent's previous message without quoting it in full
- [ ] Build clean

PR #934, opened 2026-05-17
feat(i18n): drop turn_result.annotations_heading (#596)

decay256:
Follow-up to decay256/pinder-web#633.

Frontend #596 dropped the outer `<TurnAnnotations>` frame and the "Annotations" section heading along with it. The four annotation events (triple / combo / tell / callback) now render as sibling EventBox nodes without a wrapping heading, so this i18n key has no remaining consumer in pinder-web.

Frontend stops referencing the key in pinder-web#633; this commit removes it from the upstream YAML so it doesn't sit orphaned. A comment in place of the removed line documents the migration for future spelunkers.

Refs decay256/pinder-web#596

### DIFF
--- a/data/i18n/en/ui-turn-result.yaml
+++ b/data/i18n/en/ui-turn-result.yaml
@@ -3,7 +3,10 @@
 schema_version: 1
 locale: en
 strings:
-  turn_result.annotations_heading: "Annotations"
+  # #596: turn_result.annotations_heading removed — the outer
+  # <TurnAnnotations> frame was dropped and the four annotation events
+  # (triple / combo / tell / callback) now render as sibling EventBox
+  # nodes without a section heading.
   turn_result.steering_heading: "🧭 Steering"
   turn_result.steering_explainer: "Steering: an extra roll that lets the player nudge the conversation by adding a follow-up question."
   turn_result.vs_dc: "vs DC"


decay256 (commented):
**Verdict: APPROVE**

Trivial single-file i18n cleanup. All checks pass:

- **Diff scope**: exactly one file (`data/i18n/en/ui-turn-result.yaml`), -1 / +4 lines. The added lines are a comment block in place of the removed key. Confirmed via `git diff origin/main..HEAD --stat`.
- **Orphan status**: `grep -rn annotations_heading` across the worktree returns only the new comment block in the YAML itself — no remaining consumer in pinder-core. Cross-repo check in pinder-web `origin/main` (production frontend, excluding generated `en.ts` and tests) returns empty, confirming pinder-web#633 removed the last consumer before this PR.
- **YAML well-formedness**: `yaml.safe_load` succeeds; 41 strings remain.
- **Build**: `dotnet build Pinder.Core.sln -c Release` → 0 Error(s), 191 Warning(s) (all pre-existing nullable/xUnit warnings unrelated to this PR).
- **Comment text**: accurately describes the pinder-web#633 migration — outer `<TurnAnnotations>` frame dropped, four annotation events (triple / combo / tell / callback) now render as sibling EventBox nodes without a section heading. Attribution to `#596` is correct.

No findings. Clear to merge.

Issue #942, opened 2026-05-17
[bug][P0] GameEndedException(Ghosted) in turn-prefetch state-mutates session; sync fallback then runs a phantom successful turn with negative interest


decay256:
## Symptom

In staging session `b79b6331-6698-466e-9fb1-8d859b7cb515` (pinder_staging, played 2026-05-17 17:50-17:55 UTC), the user saw turn 3 render as a successful roll (`die_roll=17, total=29, dc=28, is_success=true, miss_margin=0`) but with **interest dropping from 6 → 3** (`delta_from_roll: -3`). Successes should never produce negative `delta_from_roll`.

## Game-api logs

Two consecutive warnings logged for turn 3 BEFORE the user picked an option:

```
WARN  prefetch_faulted session=b79b6331-6698-466e-9fb1-8d859b7cb515 turn=3 elapsed_ms=4 (will fall back on next GET)
Exception: Pinder.Core.Conversation.GameEndedException: Game has ended with outcome: Ghosted
   at Pinder.Core.Conversation.GameSession.StartTurnAsync (GameSession.cs:line 791)
   at Pinder.GameApi.Services.ActiveSession.<...>.<<ResolveTurnAsync>b__2>d.MoveNext()
       (ActiveSession.cs:line 1848)

WARN  prefetch_faulted_on_drain session=b79b6331-6698-466e-9fb1-8d859b7cb515 turn=3
       ... falling back to synchronous StartTurnAsync.
       (same stack)
```

The synchronous-fallback path then caught the second throw and ran turn 3 anyway. Result: turn 3 was persisted with `is_success=true` but `interest.delta_from_roll=-3`, AND the user was never told the game had reached Ghosted.

## Why this happens (hypothesis)

`GameSession.StartTurnAsync` (`pinder-core/src/Pinder.Core/Conversation/GameSession.cs:791`) computes the start-of-turn outcome check (ghost-decay/Bored→Ghosted transition) at the top. When the speculative prefetcher in `ActiveSession.cs:1848` invokes it, the throw exits but state mutations done before the throw (notably the interest decay that pushed the session past Ghosted) appear to persist. The synchronous-fallback path (`ActiveSession.cs:1402` ish) then re-enters `StartTurnAsync` on an already-mutated session, somehow gets past the outcome check (or the ghost-decay is applied a second time and bottoms out), and produces a turn-3 record that has the bad delta.

Repro candidates:
1. Prefetch state mutation isn't being reset before the synchronous fallback.
[...]

PR #955, opened 2026-05-18
fix(#942): make StartTurnAsync transactional; add ResolveTurnAsync invariant guard

decay256:
## Summary

Fixes the core half of the P0 state-corruption bug (#942).

### Root cause

`StartTurnAsync` mutated `_ended`, `_outcome`, and `_playerShadows` (Dread +1) **before** throwing `GameEndedException` in the ghost-trigger, IsZero, and IsMaxed paths. These mutations persisted on the session object. The web-side speculative prefetch caught the exception but left the session in a half-mutated state; the synchronous fallback then re-entered the session and produced a phantom turn with `is_success=true` but `delta_from_roll=-3`.

### Fix

**Strategy B (reorder)** — throw before any state mutation:

- `StartTurnAsync` now throws `GameEndedException` without setting `_ended`, `_outcome`, or mutating `_playerShadows`. Shadow growth event descriptions are computed inline and carried in the exception (`ShadowGrowthEvents`) for the SPA to display, but the tracker is untouched.
- **Caller contract**: after catching `GameEndedException` from `StartTurnAsync`, callers must call `session.MarkEnded(ex.Outcome)`. The web PR will enforce this on the prefetch/drain-fallback path.
- Added `InvariantViolationException`: thrown by `ResolveTurnAsync` when `roll.IsSuccess == true` but `baseInterestDelta < 0` — a state that cannot occur in legitimate gameplay (SuccessScale returns ≥ 1 for success). This turns the phantom-turn scenario into a loud failure instead of silent data corruption.

### Tests

- `Issue942_StartTurnTransactionalTests.cs` — AC1 reverse-verification (ghost trigger leaves session state unchanged) + AC2 invariant guard.
- Updated `ShadowGrowthSpecTests.AC1_Ghosted_GrowsDread1`, `ShadowGrowthEventTests.Ghost_GrowsDread`, `FullConversationIntegrationTest.GameAlreadyEnded_SubsequentCallThrows` to match new transactional contract.
- Full suite: 2787 passed, 18 skipped, 0 failed.

## DoD Evidence

**Build**: `dotnet build -c Release` → 0 errors, 89 warnings (pre-existing)
**Tests targeted**: 180 passed, 1 skipped, 0 failed
**Tests full**: 2787 passed, 18 skipped, 0 failed

Partially addresses #942 (core engine side; web-side prefetch fix tracked separately and required before this issue closes).

### DIFF
--- a/src/Pinder.Core/Conversation/GameSession.cs
+++ b/src/Pinder.Core/Conversation/GameSession.cs
@@ -753,15 +753,21 @@ namespace Pinder.Core.Conversation
                 throw new GameEndedException(_outcome!.Value);
 
             // Check end conditions: interest at 0 or 25
+            //
+            // Issue #942: StartTurnAsync must be transactional — no observable state
+            // mutation (_ended, _outcome, shadow) when it throws GameEndedException.
+            // Callers (e.g. web prefetch/drain fallback) catch the exception and call
+            // session.MarkEnded(ex.Outcome) explicitly; they must not observe partial
+            // state from the failed call.
+            //
+            // Shadow growth events are still included in the exception for the SPA to
+            // display, but are NOT committed to _playerShadows before the throw.
             if (_interest.IsZero)
             {
-                _ended = true;
-                _outcome = GameOutcome.Unmatched;
-                // End-of-game Dread +1: conversation ended without date
+                // End-of-game Dread +1 event: compute description WITHOUT applying to tracker.
                 if (_playerShadows != null)
                 {
-                    _playerShadows.ApplyGrowth(ShadowStatType.Dread, 1, "Conversation ended without date");
-                    var dreadEvents = _playerShadows.DrainGrowthEvents();
+                    var dreadEvents = new[] { $"{ShadowStatType.Dread} +1 (Conversation ended without date)" };
                     throw new GameEndedException(GameOutcome.Unmatched, dreadEvents);
                 }
                 throw new GameEndedException(GameOutcome.Unmatched);
@@ -769,8 +775,6 @@ namespace Pinder.Core.Conversation
 
             if (_interest.IsMaxed)
             {
-                _ended = true;
-                _outcome = GameOutcome.DateSecured;
                 throw new GameEndedException(GameOutcome.DateSecured);
             }
 
@@ -780,14 +784,12 @@ namespace Pinder.Core.Conversation
                 int ghostRoll = _dice.Roll(4);
                 if (ghostRoll == 1)
                 {
-                    _ended = true;
-                    _outcome = GameOutcome.Ghosted;
-
-                    // Shadow growth: Ghosted → +1 Dread (#44 trigger 8)
+                    // Shadow growth: Ghosted → +1 Dread (#44 trigger 8).
+                    // Event description is surfaced via the exception for SPA display;
+                    // tracker state is NOT mutated (transactional — #942).
                     if (_playerShadows != null)
                     {
-                        _playerShadows.ApplyGrowth(ShadowStatType.Dread, 1, "Ghosted");
-                        var events = _playerShadows.DrainGrowthEvents();
+                        var events = new[] { $"{ShadowStatType.Dread} +1 (Ghosted)" };
                         throw new GameEndedException(GameOutcome.Ghosted, events);
                     }
 
@@ -1871,6 +1873,19 @@ namespace Pinder.Core.Conversation
             _currentDicePools = null;
 
             // 15. Build result
+
+            // Issue #942 invariant guard: a successful roll MUST NOT produce a negative
+            // base interest delta. SuccessScale always returns ≥ 1 for success; a negative
+            // baseInterestDelta here means the session state was pre-corrupted (e.g. by a
+            // phantom turn from the prefetch/fallback bug). Fail loudly rather than persist
+            // an incoherent turn record.
+            if (rollResult.IsSuccess && baseInterestDelta < 0)
+                throw new InvariantViolationException(
+                    $"#942 invariant violated on turn {_turnNumber}: roll.IsSuccess=true " +
+                    $"but baseInterestDelta={baseInterestDelta} (expected ≥0). " +
+                    "SuccessScale cannot produce a negative delta for a success roll. " +
+                    "This indicates a phantom turn produced from a pre-corrupted session state.");
+
             var stateSnapshot = CreateSnapshot();
 
             return new TurnResult(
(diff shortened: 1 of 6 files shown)


decay256 (commented):
## Code review — PR #955 (P0 #942 transactional fix)

**Verdict: APPROVE.** Fix is correct for scope; deferred caller-side work is documented in PR body.

### What I checked

1. **AC1 (transactional StartTurnAsync).** All three throw sites (`_interest.IsZero`, `_interest.IsMaxed`, ghost trigger) verified clean: no assignment to `_ended`/`_outcome`/`_interest`/`_turnNumber`/`_activeTraps`/`_playerShadows` before throw. Shadow event strings are constructed inline (not pulled from `_playerShadows.GrowthEvents`). `Issue942_StartTurnTransactionalTests.StartTurnAsync_ThrowsGameEnded_LeavesSessionStateUnchanged` snapshots interest + turnNumber + activeTraps + shadow effective values for all six ShadowStatTypes before and after the throw — covers all four required fields plus more.
2. **AC2 (invariant guard).** Placed at the persist boundary in `ResolveTurnAsync` after `baseInterestDelta` is computed but before the turn record is built. Implementer chose `throw InvariantViolationException` over log+reject — spawn task says non-blocking style note. Test (`NegativeSuccessDeltaRuleResolver`) constructs a synthetic violating state and asserts the throw fires. ✓
3. **Reverse-verification test.** Present, named per AC, captures correct snapshot (interest, turnNumber, activeTraps, all six shadow stat effective values). ✓
4. **Scope.** `git diff HEAD~1 HEAD --name-only -- src/` returns only `GameSession.cs` and the new `InvariantViolationException.cs`. ✓
5. **Self-verify.** `dotnet build -c Release` → 0 errors, 191 warnings (pre-existing). `dotnet test --filter "FullyQualifiedName~Issue942"` → 3 passed / 0 failed. Full suite → 2787 passed / 18 skipped / 0 failed.
6. **Web-side AC absence.** PR body explicitly defers `ActiveSession.ResolveTurnAsync` sync-fallback fix to a follow-up. Per spawn instructions, not a blocker. ✓

### The three updated pre-existing tests — verdict per file

- **`FullConversationIntegrationTest.cs` (`GameAlreadyEnded_SubsequentCallThrows`) — bug-encoded.** The test originally relied on `StartTurnAsync` itself setting `_ended=true` during the ghost throw, so the second call would also throw. That coupling was part of the bug. The update adds an explicit `session.MarkEnded(ghostEx.Outcome)` between calls, which correctly exercises the new caller contract. The test's actual intent ("an ended session refuses further StartTurnAsync calls") is preserved.
- **`ShadowGrowthEventTests.cs` (`Ghost_GrowsDread`) — bug-encoded.** Original asserted `shadows.GetDelta(Dread) == 1` after the throw, which is exactly the in-transaction mutation that caused #942. The new assertion (`== 0`) plus the preserved `ex.ShadowGrowthEvents` check correctly encodes the transactional contract.
- **`ShadowGrowthSpecTests.cs` (`AC1_Ghosted_GrowsDread1`) — mixed.** This is the load-bearing case. Test is named after spec AC#44 trigger 8 ("Ghosted → +1 Dread"). The bug part (tracker mutated mid-throw) is removed, which is correct. But the spec contract ("Ghosted causes Dread to grow by 1") is now silently downgraded to "Ghosted emits a Dread-growth event string." Until a caller (the deferred web PR, or somewhere else) actually applies the growth from `ex.ShadowGrowthEvents`, the Ghosted → Dread growth is **never applied to the tracker**. The PR is defensible because (a) the PR body explicitly commits to the web-side enforcing this and (b) the SPA-visible event payload is unchanged. But this is a contract responsibility that has been moved out of core without yet being landed elsewhere. **Non-blocking with an explicit follow-up reminder**: do not close #942 until the web PR demonstrably calls something that increments Dread on the catch path.

### Non-blocking suggestions (4)

1. **`Wait()` is now inconsistent with `StartTurnAsync`.** `CheckInterestEndConditions()` (line 2061) and `CheckGhostTrigger()` (line 2088) still set `_ended`, `_outcome`, and call `_playerShadows.ApplyGrowth(...)` BEFORE throwing `GameEndedException`. Same exception class, same outcomes, opposite transactional contract depending on which method the caller invoked. File a follow-up to make `Wait()` transactional too, or document the asymmetry loudly.
2. **`GameEndedException.ShadowGrowthEvents` is `IReadOnlyList<string>`.** Forces the web-side follow-up to parse strings (`"Dread +1 (Ghosted)"`) to re-apply growth on the caller side. Recommend adding a structured `IReadOnlyList<(ShadowStatType Stat, int Amount, string Reason)>` (or similar record) alongside the string list so callers can mechanically reapply without regex. The string form can be retained for SPA display.
3. **Invariant guard chose `throw` over `log+reject`.** Spawn task allowed either; flagging that `throw InvariantViolationException` will surface as a 500 to the user, which is loud-but-blunt. A `log+reject` path that drops the bad turn record and returns a controlled error would be more reversible. Not a blocker.
4. **PR body commitment to web PR.** The "follow-up PR will enforce caller contract" wording is correct but informal. Recommend opening the web ticket NOW (it's not in the current sprint scope) and linking it in #942 so the deferred work isn't lost.

### Blockers

None.

### Self-verify
- Build: 0 errors / 191 warnings (pre-existing).
- Issue942 tests: 3 passed / 0 failed.
- Full suite: 2787 passed / 18 skipped / 0 failed.

---

```
## Review verdict
verdict: APPROVE
blockers: 0
non_blocking: 4

## Blockers
- none

## Non-blocking suggestions
- Wait()/CheckGhostTrigger()/CheckInterestEndConditions() still mutate-before-throw; inconsistent with StartTurnAsync. File follow-up.
- GameEndedException.ShadowGrowthEvents is IReadOnlyList<string>; add a structured record list so callers don't have to parse strings to reapply growth on the catch path.
- Invariant guard uses throw rather than log+reject; spawn allowed either, log+reject is more reversible.
- File the web-side follow-up ticket now and link it from #942 so the deferred caller-contract work doesn't get lost.

## The three updated tests — verdict per file
- FullConversationIntegrationTest.cs: bug-encoded — relied on in-transaction _ended mutation; update correctly exercises new MarkEnded caller contract.
- ShadowGrowthEventTests.cs: bug-encoded — asserted tracker delta=1 mid-throw, which was the bug; new assertion correctly verifies tracker is untouched while exception still carries the event.
- ShadowGrowthSpecTests.cs: mixed — bug portion removed cleanly, but spec contract "Ghosted → +1 Dread" is now deferred to caller-side and not yet applied anywhere; non-blocking on the understanding that the web PR will land before #942 is closed.

## Self-verify
- Build: 0 errors, 191 warnings (pre-existing)
- Issue942 tests: 3 passed / 0 failed
- Full suite: 2787 passed / 18 skipped / 0 failed
```

PR #960, opened 2026-05-18
fix(#951): opening message uses opponent name, not literal 'scene'

### LINKED ISSUE #951: [bug][P1] Opening message contains literal 'scene' instead of the character name
## User report

> the opening comments say 'scene' but it needs to have the character name

## What's happening

Daniel observed an opening message rendered with the word 'scene' where the opponent's character name should be. Likely a literal template-string fragment leaking through (e.g. the prompt or i18n contains `{scene}` or a literal placeholder `SCENE`) instead of being substituted with the opponent's name.

## Acceptance criteria

- Audit `data/prompts/templates.yaml` and the opening-message generation path for any literal 'scene' / '{scene}' / '\$scene' that should be a character-name substitution.
- Replace with the runtime substitution (typically `{opponent_name}` or the canonical placeholder used elsewhere).
- Add a regression test: opening message for a known opponent should contain the opponent's name, not the literal string 'scene'.

## Repro

- Daniel observed during the 2026-05-17 staging test. Specific session/turn not yet captured — likely the opening message for either velvet or sable.
- Check the most recent opening messages by querying the opponent's first response in turn 1: `SELECT payload->'opponent'->'message' FROM turn_records WHERE turn_number = 1 ORDER BY created_at DESC LIMIT 5;`

## Anchors

- `pinder-core/data/prompts/templates.yaml` (opening / engine-opponent-block).
- `pinder-core/src/Pinder.Core/Conversation/OpponentContext.cs` (where character_name is substituted into prompts).
- `pinder-core/src/Pinder.LlmAdapters/SessionDocumentBuilder.cs` (where placeholders are resolved).

decay256:
Closes #951.

## Root cause
`SessionDocumentBuilder.AppendConversationHistory` and `BuildInterestChangeBeatPrompt` did not filter turn-0 `[scene]` entries (sender == `Senders.Scene = "[scene]"`). When a caller passed an unfiltered conversation history, scene entries appeared as `[OPPONENT|[scene]]` (in the engine block) or `[OPPONENT]` (in the beat block), giving the LLM the impression the opponent's name is scene and causing it to use that word in the opening message.

## Fix
- `AppendConversationHistory`: skip any entry where `Senders.IsScene(entry.Sender)` is true; use a separate `filteredIndex` so turn numbers remain correct for real conversational entries.
- `BuildInterestChangeBeatPrompt`: pre-filter scene entries from the conversation list before taking the last-6 window.

## Tests
- `BuildOpponentPrompt_SceneEntriesInHistory_AreFiltered` — history with scene entries → prompt contains opponent name, not `[scene]` or standalone `scene`.
- `BuildInterestChangeBeatPrompt_SceneEntriesInHistory_AreFiltered` — same guard for the beat prompt.

## DoD
- Build: clean (0 errors, 186 pre-existing warnings)
- Tests: 1069/1069 pass

### DIFF
--- a/src/Pinder.LlmAdapters/SessionDocumentBuilder.cs
+++ b/src/Pinder.LlmAdapters/SessionDocumentBuilder.cs
@@ -439,10 +439,17 @@ namespace Pinder.LlmAdapters
             if (conversationHistory != null && conversationHistory.Count > 0)
             {
                 sb.AppendLine("RECENT CONVERSATION (for context — reference specific details in your response):");
-                int start = Math.Max(0, conversationHistory.Count - 6);
-                for (int i = start; i < conversationHistory.Count; i++)
+                // #951: skip scene entries (sender == "[scene]") before computing
+                // the last-6 window so turn-0 bios and outfit descriptions never
+                // appear as [OPPONENT] lines that confuse the LLM about the
+                // opponent's character name.
+                var filtered = new System.Collections.Generic.List<(string Sender, string Text)>(conversationHistory.Count);
+                foreach (var e in conversationHistory)
+                    if (!Senders.IsScene(e.Sender)) filtered.Add(e);
+                int start = Math.Max(0, filtered.Count - 6);
+                for (int i = start; i < filtered.Count; i++)
                 {
-                    var entry = conversationHistory[i];
+                    var entry = filtered[i];
                     string role = (!string.IsNullOrEmpty(playerName) && entry.Sender == playerName) ? "PLAYER" : "OPPONENT";
                     sb.AppendLine($"[{role}] \"{entry.Text}\"");
                 }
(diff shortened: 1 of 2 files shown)


decay256 (commented):
**Verdict: APPROVE** (filing as comment — gh blocks self-approval since the PR was opened by the same bot account).

Root cause is the right one (`Senders.Scene = "[scene]"` leaking into history rendering as `[OPPONENT|[scene]]`), fix is defense-in-depth, tests are sharp.

**Correctness**
- Filter applied **before** turn numbering in `AppendConversationHistory` (uses `filteredIndex`, not raw `i`). Turn pairs stay aligned.
- Filter applied **before** the last-6 window in `BuildInterestChangeBeatPrompt`. Avoids the empty-window failure mode when last 6 are all scene tags.
- Private helper — only 3 in-file callers; no external API surface change.
- Doesn't conflict with `ConversationIndexing` / `GameSession.BuildHistoryForLlmContext` (which already filter at the source); this is a belt-and-braces guard at the render layer.

**Tests**
- Both new regression tests use `(?i)\bscene\b` whole-word match — correct (no false positives on "obscene"). Also assert opponent name present.
- 1069/1069 pass locally, build clean.

**PR hygiene**
- Body has `Closes #951`, root-cause + fix + tests clearly described.
- Commit message conventional.

No blockers. Ship it.

Correlation id: 2026-05-17-197af9-951-code-reviewer-06d349e7

Issue #206, opened 2026-03-30
AnthropicClient — HTTP client with retry logic and cache_control support


decay256:
## Context
Per architect design on issue #38. Raw HTTP client for Anthropic Messages API.

## What to build

`src/Pinder.LlmAdapters/Anthropic/AnthropicClient.cs`

```csharp
public sealed class AnthropicClient : IDisposable
{
    private const string ApiUrl = "https://api.anthropic.com/v1/messages";
    private const string AnthropicVersion = "2023-06-01";
    private const string BetaHeader = "prompt-caching-2024-07-31"; // required for cache_control
    
    private readonly HttpClient _httpClient;
    private readonly bool _ownsHttpClient;

    public AnthropicClient(string apiKey) {
        _httpClient = new HttpClient();
        _httpClient.DefaultRequestHeaders.Add("x-api-key", apiKey);
        _httpClient.DefaultRequestHeaders.Add("anthropic-version", AnthropicVersion);
        _httpClient.DefaultRequestHeaders.Add("anthropic-beta", BetaHeader);
        _ownsHttpClient = true;
    }

    public AnthropicClient(string apiKey, HttpClient httpClient) {
        _httpClient = httpClient;
        _httpClient.DefaultRequestHeaders.Add("x-api-key", apiKey);
        _httpClient.DefaultRequestHeaders.Add("anthropic-version", AnthropicVersion);
        _httpClient.DefaultRequestHeaders.Add("anthropic-beta", BetaHeader);
        _ownsHttpClient = false;
    }

    public async Task<MessagesResponse> SendMessagesAsync(MessagesRequest request, CancellationToken ct = default)
    {
        // Retry strategy:
        // 429: retry with Retry-After header, max 3 retries
        // 529: exponential backoff 1s → 2s → 4s
        // 5xx: retry once after 1s
        // 4xx (except 429): fail immediately
        // Implement retry loop here
    }

    public void Dispose() { if (_ownsHttpClient) _httpClient.Dispose(); }
}
```

## Retry policy details
- Max retries: 3
- 429: read `Retry-After` header (seconds), sleep that long; if missing, use 5s
- 529: 1s, 2s, 4s backoff
- 5xx: 1s then retry once
- 4xx (not 429, not 529): throw `AnthropicApiException` immediately
- On final failure: throw `AnthropicApiException` with status code + body

## Request serialization
- Use `JsonConvert.SerializeObject(request)` (Newtonsoft.Json)
- Content-Type: `application/json`

[...]

PR #974, opened 2026-05-18
chore(#947): wire Anthropic prompt cache on OpenRouter Sonnet-4.6 path

### LINKED ISSUE #947: [chore][P2] Anthropic prompt cache not hitting on OpenRouter — cache_read=0 every turn; input token usage grows turn-over-turn
## Symptom

Staging session `b79b6331-6698-466e-9fb1-8d859b7cb515` (`openrouter/anthropic/claude-sonnet-4.6`) shows zero prompt-cache utilisation across 3 turns:

| Turn | input_tokens | output_tokens | cache_read | cache_creation | cost_usd |
|------|--------------|---------------|------------|----------------|----------|
| 1    | 12,874       | 351           | 0          | 0              | $0.044   |
| 2    | 14,785       | 456           | 0          | 0              | $0.051   |
| 3    | 16,977       | 445           | 0          | 0              | $0.058   |

Input grows ~2k tokens per turn. Cost rises monotonically. After the #871 prompt-yaml migration (which was partly meant to make the prompt prefix-stable), cache hits were expected.

## Likely root causes

1. **Stable prefix is <1024 tokens.** Anthropic's prompt cache only registers on prefix segments ≥1024 tokens. If the per-turn dynamic content (conversation history, current state, options-asked) is interleaved with static content (system prompt, archetype directive, character sheets) rather than at the suffix, the stable prefix may not reach the threshold.

2. **Anthropic `cache_control` blocks not being set.** The Anthropic SDK requires explicit `{"type": "ephemeral"}` / `{"type": "persistent"}` cache-control markers on message blocks. If `AnthropicLlmAdapter` / `OpenAiStreamingTransport` (the OpenRouter path) isn't setting these, no cache.

3. **OpenRouter passes Anthropic prompt-cache through, but routes for Sonnet-4.6 may not.** Worth confirming with OpenRouter's docs that prompt-caching is available for `anthropic/claude-sonnet-4.6` via OpenRouter.

## Why this matters

A 20-turn session at the current growth rate would cost ~$2-3 of API spend. With prompt caching, the input cost on cached turns drops ~10x. For a fleet of staging-test + prod sessions this is meaningful.

## Acceptance criteria

- Profile turn-prompt stability: which sections are token-stable across turns vs which mutate?
- Restructure the prompt so the stable prefix is ≥1024 tokens and at the *beginning* of the message list.
- Add `cache_control: { type: \"ephemeral\" }` markers on the stable prefix blocks (Anthropic adapter path).
- Verify OpenRouter routes for Sonnet-4.6 support prompt-cache pass-through; if not, surface a config option to either route direct-Anthropic or skip the cache-control markers.
- Regression test: log-assert `cache_read_input_tokens > 0` for turn 2+ when running a 3-turn smoke session.

## Severity

**P2** — perf/cost regression, not user-blocking.

## Repro

- Session `b79b6331-6698-466e-9fb1-8d859b7cb515`, all 3 turns.
- Dump: `/tmp/staging-test-review/turn-{1,2,3}.json`.

## Anchors

- `pinder-core/src/Pinder.LlmAdapters/Anthropic/AnthropicLlmAdapter.cs`
- `pinder-core/src/Pinder.LlmAdapters/OpenAiStreamingTransport.cs` (OpenRouter path)
- Related: #871 prompt-yaml migration epic.

## Discovery

Staging-test review 2026-05-17. `/tmp/staging-test-review/findings.md`.

decay256:
Closes #947

## Diagnosis findings

- **Stable prefix token count (turn 2):** ~4-5k tokens (delivery system = 16,774 chars; opponent-response system = 17,539 chars). Both well above Sonnet-4.6's 2,048-token minimum cacheable prefix.
- **Stable prefix was interleaved:** no. Per `/tmp/staging-test-review/turn-{1,2,3}.json`, the `system_prompt` content for each phase is **byte-identical** across all three turns. The growing input tokens come entirely from the appended `user_prompt` (conversation history accumulates), not from system-prompt drift.
- **`cache_control` markers present before this PR:** no on the OpenAI-shape path (`OpenAiLlmAdapter`, `OpenAiTransport`, `OpenAiStreamingTransport`). The Anthropic-direct path (`AnthropicLlmAdapter` via `CacheBlockBuilder`) has carried the markers since #206 — staging hits that path differently.
- **OpenRouter Sonnet-4.6 `cache_control` pass-through:** confirmed-supports. Per OpenRouter docs (`/docs/guides/best-practices/prompt-caching`), Sonnet 4.6 is listed as a prompt-cache-supported model with a 2,048-token minimum, and per-block `cache_control: { type: "ephemeral" }` is the documented method for explicit breakpoints.

## DoD Evidence

- [x] Stable prefix already contiguous + ≥1024 tokens — no restructure needed.
- [x] `cache_control: ephemeral` markers now emitted on the OpenAI/OpenRouter path for the system block (the natural stable prefix). The Anthropic-direct path was already wired.
- [x] OpenRouter path: forwards markers natively for Sonnet-4.6 (and the entire Claude family per OpenRouter docs).
- [x] New regression tests (`Issue947_PromptCacheControlTests`, 6 tests, mock-based) assert the constructed payload carries the marker on both transports; disabling the gate (`UseAnthropicCacheControl=false`) reverts to plain-string content.
- [x] `dotnet build`: clean (0 errors, pre-existing warnings unchanged).
- [x] `dotnet test`:
  - `Pinder.LlmAdapters.Tests`: 1080/1080 pass (9 pre-existing skips).
  - `Pinder.Core.Tests`: 2810/2810 pass (18 pre-existing skips).
  - `Pinder.RemoteAssets.Tests`: 54/54 pass.
  - `Pinder.Rules.Tests`: 46 pre-existing failures verified against origin/main (YAML-deserialization unrelated to this change).

## Research Log

The staging-test review for session `b79b6331-...` (3 turns on `openrouter/anthropic/claude-sonnet-4.6`) showed `cache_read=0` and `cache_creation=0` on every turn while input grew 12.9→14.8→17.0k tokens — a textbook "prefix caching never engaged" pattern. The ticket flagged three plausible root causes (sub-1024-token prefix, missing `cache_control`, OpenRouter not passing markers through). Phase-A diagnosis ruled out the first and third: the per-phase `system_prompt` in the staging dumps is byte-identical across turns and 16-17k characters long (well past the 2,048-token Sonnet-4.6 minimum), and OpenRouter explicitly documents per-block `cache_control` as the supported Anthropic pass-through for Sonnet 4.6. That left root cause #2 as the entire fix.

The actual gap is narrow: `AnthropicLlmAdapter` has had `cache_control: ephemeral` on its system blocks since #206 (via `CacheBlockBuilder`), but the OpenAI-shape transports (`OpenAiLlmAdapter.BuildRequestJson`/`BuildStatefulRequestJson`, `OpenAiTransport.SendAsync`, `OpenAiStreamingTransport.SendStreamAsync`) all emit the system prompt as a plain string. Anthropic's API requires `cache_control` to live on a content-block object — plain string content cannot carry the marker. The fix wraps the system prompt in a single-element `text` content-block array with the marker inline. OpenAI native accepts the content-block shape and ignores `cache_control` (automatic caching covers them anyway); OpenRouter forwards the marker verbatim to Anthropic.

A `OpenAiOptions.UseAnthropicCacheControl` flag (default `true`) is the config gate the ticket asked for. If a future route appears that rejects unknown content-block fields, set it to `false` per-adapter without touching message structure. The new helper `OpenAiCacheControl.BuildSystemContent` is one place to find the toggle and is internal-only — call sites are the three transport entry points.

Deferred / out of scope for this PR: a full end-to-end `cache_read_input_tokens > 0` smoke test requires a real Anthropic or OpenRouter key and a multi-turn live session, which the ticket explicitly calls out as out-of-scope for a chore-level fix. Mock-based regression tests cover the request-side shape, which is the precondition the upstream cache layer keys off. Next sprint's staging-test review should confirm `cache_read` is non-zero on turn 2+ of a real session — that's the definitive landing signal.

### DIFF
--- a/src/Pinder.LlmAdapters/OpenAi/OpenAiCacheControl.cs
+++ b/src/Pinder.LlmAdapters/OpenAi/OpenAiCacheControl.cs
@@ -1,0 +1,66 @@
+using System;
+
+namespace Pinder.LlmAdapters.OpenAi
+{
+    /// <summary>
+    /// Builds the <c>content</c> field of an OpenAI-compatible
+    /// <c>messages[i]</c> entry in one of two shapes:
+    /// <list type="bullet">
+    ///   <item><description>Plain string — the legacy shape, used when prompt
+    ///   caching is disabled or when the upstream provider does not honour
+    ///   Anthropic-style <c>cache_control</c> markers.</description></item>
+    ///   <item><description>Single-element <c>text</c> content-block array with
+    ///   an inline <c>cache_control: { type: "ephemeral" }</c> breakpoint — the
+    ///   shape required by Anthropic prompt caching and by OpenRouter when
+    ///   routing to Anthropic-family models.</description></item>
+    /// </list>
+    ///
+    /// <para>
+    /// Issue #947: the OpenAI-compatible request path was sending a plain string
+    /// for the (large, byte-stable) system prompt, so the upstream Anthropic
+    /// cache layer never saw a breakpoint and <c>cache_read_input_tokens</c>
+    /// stayed at zero across multi-turn sessions. Wrapping the system prompt in
+    /// a content block with <c>cache_control: ephemeral</c> makes the prefix
+    /// cacheable on the Anthropic / OpenRouter→Anthropic path while remaining
+    /// a no-op (OpenAI accepts content-block arrays but ignores unknown fields)
+    /// on the OpenAI native path, which uses automatic caching anyway.
+    /// </para>
+    ///
+    /// <para>
+    /// OpenRouter explicitly documents per-block <c>cache_control</c> as the
+    /// supported way to enable Anthropic prompt caching for Claude models on
+    /// their platform (see
+    /// <c>https://openrouter.ai/docs/guides/best-practices/prompt-caching</c>).
+    /// Sonnet-4.6 requires a stable prefix of ≥2,048 tokens; the assembled
+    /// session system prompt is well above that threshold.
+    /// </para>
+    /// </summary>
+    internal static class OpenAiCacheControl
+    {
+        /// <summary>
+        /// Build the <c>content</c> value for a system message.
+        /// </summary>
+        /// <param name="systemPrompt">The system prompt text.</param>
+        /// <param name="useCacheControl">
+        /// When true, return a single-element content-block array with
+        /// <c>cache_control: ephemeral</c>. When false, return the plain string.
+        /// </param>
+        public static object BuildSystemContent(string systemPrompt, bool useCacheControl)
+        {
+            if (systemPrompt == null) throw new ArgumentNullException(nameof(systemPrompt));
+
+            if (!useCacheControl)
+                return systemPrompt;
+
+            return new object[]
+            {
+                new
+                {
+                    type = "text",
+                    text = systemPrompt,
+                    cache_control = new { type = "ephemeral" }
+                }
+            };
+        }
+    }
+}
--- a/src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.cs
+++ b/src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.cs
@@ -212,14 +212,21 @@ namespace Pinder.LlmAdapters.OpenAi
 
         private string BuildRequestJson(string systemPrompt, string userContent, double temperature)
         {
+            // #947: wrap the (byte-stable, large) system prompt in a content
+            // block with cache_control: ephemeral so the Anthropic /
+            // OpenRouter→Anthropic path can register a prompt-cache breakpoint.
+            // OpenAI accepts the content-block shape and ignores the marker.
+            var systemContent = OpenAiCacheControl.BuildSystemContent(
+                systemPrompt, _options.UseAnthropicCacheControl);
+
             var request = new
             {
                 model = _options.Model,
                 max_tokens = _options.MaxTokens,
                 temperature = temperature,
-                messages = new[]
+                messages = new object[]
                 {
-                    new { role = "system", content = systemPrompt },
+                    new { role = "system", content = systemContent },
                     new { role = "user", content = userContent }
                 }
             };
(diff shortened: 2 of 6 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

### Blocking
- (none)

### Non-blocking
- `OpenAiCacheControl.BuildSystemContent`: the no-op anonymous-object-array allocation per call is fine for prompt-cache hot path, but if profiling ever flags it, a static `object[]`+pre-built block can replace it — follow-up only.
- Direct-OpenAI risk surface: the wrapped content-block shape with an inline `cache_control` field is benign on OpenAI today (unknown fields ignored), but OpenAI has previously tightened request validation. The `UseAnthropicCacheControl=false` toggle is the documented escape hatch — acceptable. Worth a follow-up ticket to set it false on the direct-OpenAI `OpenAiOptions` factory if/when we ever route there.

### Positive notes
- Diagnosis methodology was sound: byte-identical `system_prompt` across 3 turns of the staging dump is the right way to rule out template drift (AC1/AC2).
- All four OpenAI-shape call sites are covered: `OpenAiLlmAdapter.BuildRequestJson` + `BuildStatefulRequestJson`, `OpenAiTransport.SendAsync`, `OpenAiStreamingTransport.SendStreamAsync` (AC3).
- `AnthropicLlmAdapter` is unchanged — its #206 `CacheBlockBuilder` path was already correct (AC3).
- AC4 verified independently: OpenRouter docs (`/docs/guides/best-practices/prompt-caching`) document per-block `cache_control: { type: "ephemeral" }` as the supported Anthropic-pass-through mechanism for Claude Sonnet family models, 2,048-token minimum.
- 6 regression tests pin both shapes: wrapped content-block array WITH `cache_control` (default), AND plain-string fallback when `UseAnthropicCacheControl=false` (AC5). Both `OpenAiTransport` and `OpenAiStreamingTransport` covered.
- Scope clean: 6 files, +349/-11 — exactly matches the spawn template's expectation. No prompt-template edits, no `AnthropicLlmAdapter` edits.
- 46 `Pinder.Rules.Tests` failures independently verified pre-existing on `origin/main` (YAML-deserialization, unrelated).
- Build clean (0 errors); `Pinder.Core.Tests` 2810/2810 pass; `Pinder.LlmAdapters.Tests` 1080/1080 pass.
- The `UseAnthropicCacheControl` toggle name and `OpenAiOptions` location follow existing convention; default `true` is correct.
- `OpenAiCacheControl.BuildSystemContent` is a clean, internal-only static helper — no DI, no over-engineering.

**Verdict: APPROVE** (self-approve blocked; --comment fallback per SELF-APPROVE-BLOCKED)

Last impl of the sprint. Diagnosis-led; minimal blast radius; properly gated; regression-tested at the request-shape level (the only level reachable without live API keys, which the ticket explicitly scoped out). Confirmation of the actual `cache_read_input_tokens > 0` signal is next sprint's staging-test review.

Issue #976, opened 2026-05-20
[#964 follow-up] Engine-side population of RollCheckResult/ShadowCheckResult/HornyCheckResult.Consequence from i18n catalogue


decay256:
## Parent
#964 — wire half delivered. `Consequence` property + `ApplyConsequence` now exists on all three wire DTOs but is always null at runtime.

## Scope
- Load a consequence catalogue from `data/i18n/en/consequences.yaml` (or similar path structure)
- Implement slot substitution for consequence template strings
- Call `ApplyConsequence` from `RollEngine`, `ShadowCheckEngine`, and `HorninessCheckEngine` at appropriate points
- Handle locale selection / i18n loader wiring

## Note
`TropeTrapResult` does not exist as a separate class. `TropeTrap` is a `FailureTier`, captured on `RollCheckResult` when `Tier == FailureTier.TropeTrap`. The `Consequence` field on `RollCheckResult` covers the TropeTrap case.

PR #977, opened 2026-05-20
chore(#964): nullable consequence wire field on RollCheckResult / ShadowCheckResult / HorninessCheckResult

### LINKED ISSUE #964: [chore][P2] [#649 follow-up] Surface 'consequence' on RollCheckResult / ShadowCheckResult / HornyCheckResult / TropeTrapResult wire DTOs
## Background

pinder-web #649 added plain-language consequence strings to the expanded EventBox using Path B (SPA i18n catalogue). The SPA now has per-kind/per-tier strings from `data/i18n/en/consequences.yaml` resolved client-side.

## Problem

The i18n fallback approach (Path B) works but has limitations:
- Consequence text is static per tier, not aware of dynamic context (e.g. which specific trap variant fired, exact roll margin, opponent-state-sensitive phrasing).
- Engine knows the consequence at event time; surfacing it on the wire DTO would be the single source of truth.

## Proposed change

Add `consequence: string` field to:
- `RollCheckResult` — filled when `IsSuccess == false` or on Nat 20.
- `ShadowCheckResult` — filled when `IsMiss == true`.
- `HornyCheckResult` — filled when `Tier != None`.
- `TropeTrapResult` — always filled (the trap consequence is always user-visible).

The SPA would prefer the wire field when present, fall back to i18n catalogue when absent (graceful upgrade path — no SPA breakage on old sessions).

## Scope

Engine side only. SPA wire-up is trivial once the field exists (pass through `rollAdapters.ts`, render via the existing `plainConsequence` prop added in #649).

This is Path A from the original #649 ticket body.

decay256:
Closes #964

Adds nullable `Consequence` property (`string?`, default `null`) with
`[JsonPropertyName("consequence")]` on `RollCheckResult`, `ShadowCheckResult`,
and `HorninessCheckResult`. Each class receives an `ApplyConsequence(string)`
method with an idempotent guard — second call throws `InvalidOperationException`.
SPA falls back to client-side i18n catalogue when `consequence` is null.

## Out of scope
- Engine-side population (loading yaml catalogue, slot substitution, locale handling) → follow-up **#976**
- `TropeTrapResult`: no separate class exists. `TropeTrap` is a `FailureTier`, captured on
  `RollCheckResult` when `Tier == FailureTier.TropeTrap`. The `Consequence` field on
  `RollCheckResult` covers the TropeTrap case.

## DoD Evidence
- `dotnet build` clean (0 errors)
- `dotnet test` — all 9 new tests pass (3 per class: default-null, setter, idempotent guard)
- `consequence` key serialises as snake_case via `[JsonPropertyName]`
- Existing tests unaffected

## Research Log
- Audited `FailureTier` enum values before writing tests (initially used non-existent `Near`; corrected to `Fumble`)
- Confirmed `TropeTrapResult` does not exist in the codebase — `TropeTrap` is a `FailureTier` value, not a separate result class

### DIFF
--- a/src/Pinder.Core/Conversation/HorninessCheckResult.cs
+++ b/src/Pinder.Core/Conversation/HorninessCheckResult.cs
@@ -1,3 +1,4 @@
+using System.Text.Json.Serialization;
 using Pinder.Core.Rolls;
 
 namespace Pinder.Core.Conversation
@@ -35,6 +36,24 @@ namespace Pinder.Core.Conversation
         /// Null only for the <see cref="NotPerformed"/> sentinel.
         /// </summary>
         public RollCheckResult? Check { get; }
+
+        /// <summary>
+        /// Human-readable consequence text for this horniness check (#964).
+        /// Population deferred to follow-up. SPA falls back to client-side i18n catalogue when null.
+        /// </summary>
+        [JsonPropertyName("consequence")]
+        public string? Consequence { get; private set; }
+
+        /// <summary>
+        /// Apply an engine-populated consequence string. Idempotent — throws
+        /// <see cref="System.InvalidOperationException"/> if already set (#964).
+        /// </summary>
+        public void ApplyConsequence(string consequence)
+        {
+            if (Consequence != null)
+                throw new System.InvalidOperationException("Consequence already applied");
+            Consequence = consequence;
+        }
 
         public HorninessCheckResult(
             int roll, int modifier, int total, int dc,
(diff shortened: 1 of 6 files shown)


decay256 (commented):
## Code Review (comment-mode approval — gh blocks self-approve)

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 0

### Blocking
None.

### Non-blocking
None.

### Positive notes
- All 3 result classes carry nullable `Consequence` with `[JsonPropertyName("consequence")]`.
- `ApplyConsequence(string)` idempotent guard implemented identically across all 3 — throws `InvalidOperationException` with consistent message.
- XML docs reference #964 and the SPA i18n fallback contract.
- 9 tests total (3 per class): default-null, setter-sets, second-call-throws. Covers behaviour not implementation.
- PR body cleanly scopes engine-side population to follow-up #976 and explains `TropeTrapResult` absence (TropeTrap is a `FailureTier` on `RollCheckResult`).
- DoD Evidence + Research Log present.
- No unrelated refactors; pure additive wire-schema change.

**Verdict**: APPROVE

PR #980, opened 2026-05-20
fix(#956): add typed ShadowGrowthEffects record list to GameEndedException

### LINKED ISSUE #956: [#942 follow-up] GameEndedException.ShadowGrowthEvents should expose a structured record list
Filed by code-reviewer on PR #955 (the core engine half of #942).

## Context

PR #955 made `StartTurnAsync` transactional: when it throws `GameEndedException(Ghosted)`, the player-shadows tracker is no longer mutated. The growth events that *would* have been applied are now carried in `GameEndedException.ShadowGrowthEvents` as `IReadOnlyList<string>` — descriptive strings like `"Dread +1 (Ghosted)"`.

The caller (the web-side fix tracked under #942, still pending) needs to apply those growth deltas after catching the exception, so that Ghosted → +1 Dread is actually recorded on the session (spec AC#44 trigger 8).

## Problem

The current shape forces the caller to regex-parse `"Dread +1 (Ghosted)"` to reconstruct `(ShadowStatType.Dread, +1, "Ghosted")`. That's brittle: format drift in `GameSession` silently breaks growth in callers.

## Proposed fix

Add a second property on `GameEndedException` carrying the same data as a typed record:

```csharp
public sealed record ShadowGrowthEffect(ShadowStatType Stat, int Amount, string Reason);

public IReadOnlyList<ShadowGrowthEffect> ShadowGrowthEffects { get; }
```

The string list `ShadowGrowthEvents` stays for SPA display. Callers reapply growth from `ShadowGrowthEffects` mechanically.

## AC

- [ ] `GameEndedException` carries `ShadowGrowthEffects` as well as `ShadowGrowthEvents`.
- [ ] All sites that currently build the string list also build the record list.
- [ ] Existing string-list consumers (SPA `shadow-growth-events` wire field) unchanged.
- [ ] Unit test asserts both lists are populated and consistent on a Ghosted throw.

## Severity

P2 — quality-of-life for the #942 web caller. The web fix can ship without this by regex-parsing, but doing both at once is cleaner.

Refs: PR #955, issue #942.

decay256:
Closes #956

Adds parallel typed list ShadowGrowthEffects: IReadOnlyList<ShadowGrowthEffect> on GameEndedException alongside the existing string ShadowGrowthEvents. Callers catching the exception (the web-side fix tracked under #942) can now reapply growth deltas mechanically instead of regex-parsing 'Dread +1 (Ghosted)'.

- New record: ShadowGrowthEffect(ShadowStatType, int Amount, string Reason)
- New constructor: GameEndedException(outcome, events, effects)
- All existing call sites in GameSession.cs updated to build both lists 1:1
- SessionShadowTracker exposes a helper for building the typed list parallel to events
- New regression test asserts both lists are populated and consistent on Ghosted

Tests: Pinder.Core.Tests 2822/0 passed (18 skipped). Build clean.

Implementer subagent stream-cut after running tests successfully; orchestrator squashed and pushed the on-disk work after build+test re-verification (zero_token_stats flake, work was complete on disk).

### DIFF
--- a/src/Pinder.Core/Conversation/GameEndedException.cs
+++ b/src/Pinder.Core/Conversation/GameEndedException.cs
@@ -1,8 +1,46 @@
 using System;
 using System.Collections.Generic;
+using Pinder.Core.Stats;
 
 namespace Pinder.Core.Conversation
 {
+    /// <summary>
+    /// Typed record for a single shadow growth event, carrying the same data as the
+    /// human-readable string in <see cref="GameEndedException.ShadowGrowthEvents"/>.
+    /// Callers that need to reapply growth (e.g. after catching a Ghosted throw)
+    /// should consume <see cref="GameEndedException.ShadowGrowthEffects"/> rather
+    /// than regex-parsing the string list.
+    /// </summary>
+    public sealed class ShadowGrowthEffect : IEquatable<ShadowGrowthEffect>
+    {
+        public ShadowStatType Stat { get; }
+        public int Amount { get; }
+        public string Reason { get; }
+
+        public ShadowGrowthEffect(ShadowStatType stat, int amount, string reason)
+        {
+            Stat = stat;
+            Amount = amount;
+            Reason = reason ?? throw new ArgumentNullException(nameof(reason));
+        }
+
+        public override bool Equals(object obj) => obj is ShadowGrowthEffect other && Equals(other);
+        public bool Equals(ShadowGrowthEffect other) =>
+            !(other is null) && Stat == other.Stat && Amount == other.Amount && Reason == other.Reason;
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hash = 17;
+                hash = hash * 31 + Stat.GetHashCode();
+                hash = hash * 31 + Amount.GetHashCode();
+                hash = hash * 31 + (Reason?.GetHashCode() ?? 0);
+                return hash;
+            }
+        }
+        public override string ToString() => $"{Stat} {Amount:+0;-#} ({Reason})";
+    }
+
     /// <summary>
     /// Thrown when an operation is attempted on a GameSession that has already ended.
     /// </summary>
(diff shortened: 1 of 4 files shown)


decay256 (commented):
**Verdict: APPROVE**

## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 2

### AC coverage (all 4)
- ✅ AC#1 — `GameEndedException.ShadowGrowthEffects` typed list added.
- ✅ AC#2 — All four production throw sites that build the string list now also build the typed list (GameSession.cs L772/L795/L2083/L2114, both directly-constructed arrays and `SessionShadowTracker.Drain*` paths).
- ✅ AC#3 — String list `ShadowGrowthEvents` untouched; `TurnResult.ShadowGrowthEvents` wire field unchanged; no caller of the existing 2-arg ctor in src/ remains (legacy 2-arg ctor preserved with empty-effects default → backwards compat OK).
- ✅ AC#4 — `GhostedException_CarriesTypedShadowGrowthEffects` asserts both lists populated, equal Count, and indexes into element[0] for Stat/Amount/Reason. Two extra ctor unit tests cover default + 3-param paths.

### API surface
- `ShadowGrowthEffect(ShadowStatType Stat, int Amount, string Reason)` shape matches ticket spec verbatim. Note: project targets `LangVersion 8.0` (records require C# 9+), so implementer used a `sealed class` with manual `IEquatable<T>`, `GetHashCode`, and `ToString` — functionally equivalent to the spec'd `record` for web-caller consumers (property access + value equality). Correct call.

### Build + tests
- `dotnet build src/Pinder.Core/` → 0 errors, 32 pre-existing nullable warnings (none introduced).
- `dotnet test tests/Pinder.Core.Tests/` → **Passed! Failed: 0, Passed: 2822, Skipped: 18** (full suite).
- ShadowGrowth-filtered run: 90/90 passing.

### Non-blocking
- [SessionShadowTracker.cs:144] In `ApplyOffset`, the floored case stores slightly different reason text in the string list vs the typed effect: string is `"{reason} (floored)"` wrapped in parens (`(reason) (floored)`) while effect reason is `"{reason} (floored)"` unparenthesized. Parallelism by index/Count is preserved; only the literal Reason text drifts. Cosmetic — typed callers won't care, SPA string-rendering unchanged.
- [Stats/SessionShadowTracker.cs:3] New `using Pinder.Core.Conversation;` creates a soft circular namespace coupling (Stats → Conversation → Stats). Compiles fine (single assembly). Could be tidied later by moving `ShadowGrowthEffect` into `Pinder.Core.Stats` since it's stats-shaped data, but the ticket spec placed it in the GameEndedException file so this follows spec. Defer.

### Positive notes
- `Clone()` and `RestoreFromSnapshot()` correctly mirror the new `_growthEffects` list — no fork-mode regressions.
- Pinder stat sanity: tests use `ShadowStatType.Dread` / `Madness` (legitimate shadow stats); no Chaos misuse.
- Test asserts `Count` equality, not just `NotEmpty`. Strong parallelism check per ticket AC.

**Verdict**: APPROVE

PR #981, opened 2026-05-20
#957 Make Wait() path helpers transactional (CheckInterestEndConditions, CheckGhostTrigger, step-8)

### LINKED ISSUE #957: [#942 follow-up] Wait() / CheckInterestEndConditions() / CheckGhostTrigger() still mutate-before-throw — inconsistent with StartTurnAsync
Filed by code-reviewer on PR #955.

## Context

PR #955 made `StartTurnAsync` transactional: throws `GameEndedException` with zero state mutation. The caller (after #942 web-side lands) is responsible for `MarkEnded` + growth application.

But the **other** code path that throws `GameEndedException` from `GameSession` — the `Wait()` family — still has the OLD mutate-then-throw pattern:

- `GameSession.cs:2054-2090` `CheckInterestEndConditions()`: sets `_outcome = GameOutcome.Unmatched` (line 2056), then throws.
- `GameSession.cs:2078-2090` `CheckGhostTrigger()`: sets `_outcome = GameOutcome.Ghosted`, calls `_playerShadows.ApplyGrowth(Dread, 1, "Ghosted")`, then throws.

Both are called from `Wait()`, which is the player's "skip a turn" path.

## Problem

Same `GameEndedException` class, same outcomes, opposite transactional contract depending on which method the caller invoked. A caller catching `GameEndedException` from `Wait()` must NOT call `session.MarkEnded(ex.Outcome)` (already set) NOR reapply shadow growth (already applied). A caller catching from `StartTurnAsync` MUST do both. This is a footgun for the web layer.

## Options

**A. Make `Wait()` transactional too.** Apply the same Strategy B reorder to `CheckInterestEndConditions` and `CheckGhostTrigger`. Caller contract becomes uniform: catch `GameEndedException` ⇒ caller applies `MarkEnded` + growth from `ex.ShadowGrowthEvents` (or `.ShadowGrowthEffects` per #956), regardless of which engine method threw. **Recommended.**

**B. Document the asymmetry loudly.** Big XMLDoc comment on `GameEndedException` explaining that `Wait()`-thrown instances are already-applied and `StartTurnAsync`-thrown instances need caller follow-through. **Not recommended** — relies on caller reading comments.

## AC

- [ ] Pick A or B.
- [ ] If A: `Wait()` path no longer mutates `_outcome` or `_playerShadows` before throw. Reverse-verification test analogous to `Issue942_StartTurnTransactionalTests` on the `Wait` path.
- [ ] If B: XMLDoc on `GameEndedException` documents which throwing method requires what caller action. `Wait()`-thrown instances may need a separate exception type or a discriminator property.

## Severity

P1 — same root cause as #942 (`GameEndedException` thrown after partial state commit) on a different entry point. The web side of #942 will catch `GameEndedException` from BOTH entry points and currently has no way to tell them apart.

Refs: PR #955, issue #942, code-review thread on #955.

decay256:
## Summary

Makes the Wait() path end-condition checks transactional — no observable state mutation (`_ended`, `_outcome`, shadow tracker) before throwing `GameEndedException`. Uniform contract with `StartTurnAsync` per `#942`.

### Changes

- **`CheckInterestEndConditions()`** — removes `_ended`/`_outcome` mutation and `_playerShadows.ApplyGrowth` before throw. Builds typed `ShadowGrowthEffect[]` literals inline.
- **`CheckGhostTrigger()`** — same treatment.
- **`Wait()` step 8** — end-condition check after interest change now throws `GameEndedException(Unmatched)` with typed shadow effects instead of direct state mutation.
- **`GameSessionWaitTests.cs`** — updated 2 tests (`Wait_InterestHitsZero_GameEnds`, `Wait_OnEndedGame_ThrowsGameEndedException`) to catch `GameEndedException` and call `session.MarkEnded(ex.Outcome)` per the new uniform contract.
- **`Issue957_WaitTransactionalTests.cs`** — reverse-verification tests: Unmatched (interest=0), Ghosted (ghost trigger), DateSecured (max interest), typed `ShadowGrowthEffects` on exception, caller-side MarkEnded contract.

### Uniform contract

A caller catching `GameEndedException` from any engine method now:
1. Calls `session.MarkEnded(ex.Outcome)`
2. Reapplies growth from `ex.ShadowGrowthEffects`

Closes #957

### DIFF
--- a/src/Pinder.Core/Conversation/GameSession.cs
+++ b/src/Pinder.Core/Conversation/GameSession.cs
@@ -2026,16 +2026,25 @@ namespace Pinder.Core.Conversation
         /// Self-contained turn action — does NOT require StartTurnAsync() first.
         /// </summary>
         /// <exception cref="GameEndedException">If the game has already ended or ghost trigger fires.</exception>
+        /// <summary>
+        /// Waits one turn: -1 interest, advance trap timers, increment turn.
+        /// </summary>
+        /// <remarks>
+        /// #957: uniform transactional contract — helpers (CheckInterestEndConditions,
+        /// CheckGhostTrigger) and step-8 check do not mutate state before throwing.
+        /// Callers catch GameEndedException and call session.MarkEnded(ex.Outcome) +
+        /// reapply shadow growth from ex.ShadowGrowthEffects.
+        /// </remarks>
         public void Wait()
         {
             // 1. Check if game already ended
             if (_ended)
                 throw new GameEndedException(_outcome!.Value);
 
-            // 2. Check interest end conditions
+            // 2. Check interest end conditions (transactional — #957)
             CheckInterestEndConditions();
 
-            // 3. Ghost trigger check
+            // 3. Ghost trigger check (transactional — #957)
             CheckGhostTrigger();
 
             // 4. Clear pending Speak options and weakness window (#49)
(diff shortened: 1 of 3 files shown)


decay256 (commented):
**Verdict: APPROVE**

## Code Review (PR #981 — #957 Wait() transactional)

### Blocking
*(none)*

### Non-blocking
- **`GameSession.cs:2026-2037`** — duplicate `<summary>` tag on `Wait()`. The pre-existing `/// <summary>Self-contained turn action — does NOT require StartTurnAsync() first.</summary>` was kept, then a second `<summary>Waits one turn: -1 interest…</summary>` was added immediately after (with `<exception>` sandwiched between). Produces malformed XMLDoc — pick one summary or merge them.
- **`Issue957_WaitTransactionalTests.cs:340-356`** — `Wait_CallerReappliesShadowGrowth_DreadIncremented` is misnamed: the body never calls `shadows.ApplyGrowth(...)` to actually reapply growth. It only asserts the effect is present in the exception. The trailing comment acknowledges this. Either rename to `…ExceptionCarriesEffectForCallerToReapply` or actually invoke the reapply path and assert `GetEffectiveShadow(Dread) == 1` afterward (the latter would also pin caller-side semantics).
- **`GameSessionWaitTests.cs:Wait_InterestHitsZero_GameEnds`** — the updated assertion checks throw + outcome + post-MarkEnded re-throw, but does not explicitly assert the *transactional* property (`IsEnded == false` / `_outcome == null` *before* MarkEnded). The new `Issue957_WaitTransactionalTests` covers this in dedicated tests, so the gap is not duplicated work — but the bug-regression test arguably lost a tooth on the specific "step-8 mutate-before-throw" path. Optional: assert `!session.IsEnded` between catch and MarkEnded.

### Positive notes
- **Mirror fidelity to StartTurnAsync is exact.** Both `CheckInterestEndConditions` (interest=0 branch) and `CheckGhostTrigger` use the identical inline `new[] { $"{ShadowStatType.Dread} +1 (…)" }` events + `new[] { new ShadowGrowthEffect(…) }` effects pattern as `StartTurnAsync` at lines 766-800. No `DrainGrowthEvents/Effects` calls. No partial mutation.
- **Step-8 extension is principled.** Original step-8 had the same mutate-then-end root cause (`_ended=true; _outcome=…; ApplyGrowth(…)`); the PR converts it to the same Option A throw. I checked all `.Wait()` callers in `src/` (zero — only tests) and the 7 test callsites (`WeaknessWindowTests`, `CritAdvantageTests`, `FullConversationIntegrationTest`, `ComboGameSessionTests`, `DecomposedModuleTests`, `XpTrackingSpecTests`, `TellBonusTests`). None drive interest to 0 via `Wait()`; none rely on silent-end behavior. The only test that did (`Wait_InterestHitsZero_GameEnds`) was updated. No production-side breakage.
- **Pinder stat hygiene.** Only `ShadowStatType.Dread` appears. Chaos correctly treated as a positive stat, not a shadow.
- **AC1-AC5 coverage.** New `Issue957_WaitTransactionalTests` covers Unmatched (interest=0), Ghosted (ghost trigger d4=1 in Bored), DateSecured (max interest), typed `ShadowGrowthEffects` payload, and the caller-side MarkEnded contract — directly mirroring `Issue942_StartTurnTransactionalTests`.
- **PR body explicitly documents the scope extension** to step-8 and the uniform caller contract. Not silent scope-bleed.

### DoD evidence
- `dotnet build` → 0 errors, 222 warnings (all pre-existing).
- `dotnet test` (full suite, `/tmp/work-957`):
  - `Pinder.Rules.Tests.dll`: 244 passed, 0 failed.
  - `Pinder.Core.Tests.dll`: 2829 passed, 18 skipped, 0 failed.
  - `Pinder.LlmAdapters.Tests.dll`: 1080 passed, 9 skipped, 0 failed.
  - **Total: 4153 passed, 0 failed.**
- Diff scope: `GameSession.cs` (+25/-23), `GameSessionWaitTests.cs` (+18/-3), new `Issue957_WaitTransactionalTests.cs` (+370/0). Three files, all in declared scope.

Approving on a `--comment` review per SELF-APPROVE-BLOCKED (gh token identity == author).

PR #984, opened 2026-05-21
[infra] Wire prompt-content grep gate as a local eigentakt phase-4 check

### LINKED ISSUE #885: [infra] Wire prompt-content grep gate as a local eigentakt phase-4 check
Follow-up from PR closing #875 (Phase 5 deliverable B of the prompt-yaml migration epic, formerly #871).

## Context update (2026-05-20)

The original framing of this ticket asked for a `.github/workflows/prompt-content-gate.yml` workflow. **That framing is invalid under pinder's CI policy** (see `~/.openclaw/agents-extra/pinder/memory/ci-policy.md`): pinder never uses GitHub workflows, GitHub Actions, or any vendor-hosted CI — AI agents run all gates locally in worktrees before merge.

The need behind the ticket is still real, just the delivery shape changes.

## What to do (revised)

Ship a local gate that blocks PRs reintroducing hardcoded prompt content in C# const strings. Two acceptable shapes:

1. **`scripts/check-prompt-content.sh`** — standalone bash script that greps for prompt-shape markers (e.g. `"Stat:"`, `"OPTION_"`, `"[SIGNALS]"`, `"ACTIVE ARCHETYPE"`, `"FUNDAMENTAL RULE"`, `"You are playing the role of"`) in `const string` declarations under `src/` outside the known-allowed paths. Exit non-zero on violations. Wire into the eigentakt orchestrator's Phase 4 checks via `.eigentakt.yaml` `test_commands` (the current schema's only hook point — see Refiner assumptions below).

2. **eigentakt phase-4 inline hook** — add the grep directly to the orchestrator's local gate sequence with a structured failure mode. Same content, lighter packaging.

Recommendation: option 1 (script under `scripts/`). Reusable, auditable, runnable outside eigentakt for ad-hoc checks.

## Known-allowed residue at time of filing

Three const-string prompts remain in `src/` and are legitimate (do not block on them):
- `Pinder.SessionSetup/LlmStakeGenerator.cs:54` `DefaultSystemPrompt` — yaml-backed fallback path retained.
- `Pinder.SessionSetup/LlmOutfitDescriber.cs:22` `SystemPrompt` — migration candidate for follow-up.
- `Pinder.Core/Conversation/PlayerResponseDelayEvaluator.cs:12` `DefaultTestPrompt` — test fixture, not a production prompt.

The gate should allowlist these three exact lines (file-path relative to repo root, pinned by path only — not line number, since surrounding context may shift) and reject any new const-string prompt elsewhere.

## AC

- [ ] `scripts/check-prompt-content.sh` exists on `main`, exits 0 on a clean tree.
- [ ] Adding a new `const string DemoPrompt = "Stat: ..."` anywhere under `src/` (outside allowlist) causes the script to exit non-zero.
- [ ] `.eigentakt.yaml` gains a `test_commands` entry running `bash scripts/check-prompt-content.sh`; the orchestrator's Phase 4 test phase thus blocks merge on non-zero exit.
- [ ] Allowlist mechanism documented in the script header (format: one relative path per line, comments allowed).
- [ ] Script is runnable standalone (`bash scripts/check-prompt-content.sh`) from the repo root without extra dependencies beyond `grep`/`bash`.

## Refiner assumptions

- **`.eigentakt.yaml` hook key:** picked `test_commands` as the wiring point. The manifest schema (`.eigentakt-manifest-schema.md` in the eigentakt skill) only defines `build_commands` and `test_commands` — there is no `pre_merge_checks` key. Adding the check to `test_commands` means it fires during the Phase 4 test step, which is the nearest equivalent to a pre-merge gate. Alternative: an inline orchestrator call outside the YAML; rejected because that requires changing the orchestrator spec, whereas YAML is self-contained and forward-compatible. Reversible by moving the entry to a future `pre_merge_checks` key if the schema gains one.
- **Allowlist format:** picked file-path-only (not file:line pinned) despite the original ticket saying "file:line pinned." Line numbers shift with normal development; pinning by path is more durable and the three known-residue files are identifiable by path alone. Reversible by switching to a line-anchored grep pattern in the script if the implementer judges path-only too broad.

## Refined-by ticket-refiner

Refined: 2026-05-21 (sprint 2026-05-21-06c60a)

decay256:
Closes #885. Implements a local gate scripts/check-prompt-content.sh that blocks PRs reintroducing hardcoded prompt content in C# const strings, with an allowlist of legacy/test prompts. Wires it in .eigentakt.yaml under test_commands.

### DIFF
--- a/.eigentakt.yaml
+++ b/.eigentakt.yaml
@@ -10,3 +10,6 @@ test_commands:
   - cmd: "dotnet test"
     transition: TESTS_RAN
     output_parser: dotnet
+  - cmd: "bash scripts/check-prompt-content.sh"
+    transition: TESTS_RAN
+
--- a/scripts/check-prompt-content.sh
+++ b/scripts/check-prompt-content.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+# ==============================================================================
+# Prompt Content Gate - scripts/check-prompt-content.sh
+#
+# This script blocks the reintroduction of hardcoded prompt content in C# const
+# strings under the src/ directory.
+#
+# Allowlist mechanism (one relative path per line, comments/empty lines allowed):
+# --- ALLOWLIST START ---
+# src/Pinder.SessionSetup/LlmStakeGenerator.cs
+# src/Pinder.SessionSetup/LlmOutfitDescriber.cs
+# src/Pinder.Core/Conversation/PlayerResponseDelayEvaluator.cs
+# --- ALLOWLIST END ---
+# ==============================================================================
+
+# Ensure we run from the repo root
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+cd "$REPO_ROOT"
+
+# Parse allowlist from the script header
+allowlist=()
+while read -r line; do
+    if [[ "$line" =~ ^#[[:space:]]+src/ ]]; then
+        clean_line=$(echo "$line" | sed -E 's/^#[[:space:]]*//' | xargs)
+        allowlist+=("$clean_line")
+    fi
+done < "$0"
+
+echo "Parsed allowlist:"
+for allowed in "${allowlist[@]}"; do
+    echo "  - $allowed"
+done
+echo ""
+
+# Find all C# files in src/
+files=$(find src/ -name "*.cs")
+
+has_violations=0
+
+for file in $files; do
+    # Check if file is in allowlist
+    is_allowed=0
+    for allowed in "${allowlist[@]}"; do
+        if [[ "$file" == "$allowed" ]]; then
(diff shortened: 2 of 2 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 1

### Blocking
- None.

### Non-blocking
- [scripts/check-prompt-content.sh:73] The AWK parser uses `/;`/ to detect the end of a C# `const string` statement. If a multi-line raw string literal contains an internal semicolon (e.g. within a description or instruction) before the final terminating semicolon, the parser will evaluate the block prematurely. This is minor and non-blocking for a simple regex-based gate, but worth keeping in mind.

### Positive notes
- Robust local gating for prompt migration compliance.
- Dynamic allowlist parsing co-locates configuration and documentation in the script header.
- Handled raw and verbatim multi-line string C# declarations correctly during verification.

**Verdict**: APPROVE

PR #995, opened 2026-05-23
refactor(RemoteAssets): split EigencoreCharacterStore into Read and SyncHelper

### LINKED ISSUE #4: Vision: Issue #1 will break test suite — test updates must ship with code changes
## Concern
Issue #1 changes `StatBlock.DefenceTable` pairings and base DC from 10 to 13. Issue #2 updates all hardcoded test assertions to match — but #2 depends on #1 being merged first.

This means merging #1 alone will leave `main` with a broken test suite: every test asserting `DC == 10` or using the old defence pairings will fail. There's a red-CI window between #1 merge and #2 merge.

## Why it matters
- Broken main blocks all other work and violates the invariant that `main` is always green.
- In an automated swarm, agents that run `dotnet test` as validation will incorrectly report failures.
- If #2 is delayed or fails review, the broken window grows.

## What needs to happen
- **Option A (preferred):** Fold the test updates from #2 into #1 so the code change and test fixes land in a single atomic PR. #2 then only handles the InterestMeter.Max change (which is independently testable).
- **Option B:** Combine #1 and #2 into a single issue/PR.
- **Acceptance:** After merging the PR that changes base DC and defence pairings, `dotnet test` passes with zero failures on main.

decay256:
Modularize EigencoreCharacterStore by splitting read/query path mechanisms from write/publish/delete pipeline actions.

Closes #4


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 0

### Blocking
None.

### Non-blocking
None.

### Positive notes
- Excellent architectural split! Separating read paths into `EigencoreCharacterStoreRead` and write/delete paths into `SyncHelper` substantially reduces file complexity while maintaining clean separation of concerns.
- `EigencoreCharacterStore.cs` remains the unified entry point and delegates seamlessly to the new helper classes.
- HttpClient management is thread-safe and respects socket-exhaustion guidance by disabling handler disposal (`disposeHandler: false`).
- 429 rate limit retry loops are perfectly encapsulated and clean.

**Verdict**: APPROVE

PR #996, opened 2026-05-23
feat: split SessionDocumentBuilder

### LINKED ISSUE #4: Vision: Issue #1 will break test suite — test updates must ship with code changes

decay256:
Extract the list formatting of historical messages into a standalone static helper class HistoryFormatter.

Closes #4

### DIFF
--- a/src/Pinder.LlmAdapters/HistoryFormatter.cs
+++ b/src/Pinder.LlmAdapters/HistoryFormatter.cs
@@ -1,0 +1,81 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Pinder.Core.Conversation;
+
+namespace Pinder.LlmAdapters
+{
+    /// <summary>
+    /// Helper for formatting conversation history for LLM prompts.
+    /// </summary>
+    public static class HistoryFormatter
+    {
+        /// <summary>
+        /// Formats conversation history with [T{n}|PLAYER|name] markers.
+        /// Full history — never truncated.
+        /// </summary>
+        public static void Format(
+            StringBuilder sb,
+            IReadOnlyList<(string Sender, string Text)> history,
+            string playerName)
+        {
+            if (sb == null) throw new ArgumentNullException(nameof(sb));
+            if (history == null) throw new ArgumentNullException(nameof(history));
+
+            sb.AppendLine("[CONVERSATION_START]");
+
+            // #951: turn index must count only real conversational entries.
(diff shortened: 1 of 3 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 1

### Blocking
_None._

### Non-blocking
- `HistoryFormatter.cs`: Null-guard tests for `ArgumentNullException` in both `Format` and `FormatRecent` are absent. The guards themselves are correct and well-placed; test coverage for them would complete the defensive contract.

### Positive notes
- `HistoryFormatter.cs` is correctly placed at `src/Pinder.LlmAdapters/HistoryFormatter.cs` in the right namespace.
- `AppendConversationHistory` is fully removed from `SessionDocumentBuilder.cs` — all 3 call sites cleanly replaced with `HistoryFormatter.Format`, the `FormatRecent` inline block replaced with `HistoryFormatter.FormatRecent`. No dead code left.
- Behavior is provably equivalent: the method bodies are direct lifts with no logic changes.
- Null guards (`ArgumentNullException`) added to both public methods — an improvement over the original private method.
- `playerName` is `string?` in `FormatRecent`, matching the original nullable-aware `!string.IsNullOrEmpty` guard.
- `Senders.IsScene` cross-namespace dependency is correctly sourced from `Pinder.Core.Conversation`.
- Tests: 2/2 new `HistoryFormatterTests` pass; full suite 1082 pass / 9 skip / 0 fail.

**Verdict**: APPROVE

PR #999, opened 2026-05-24
feat(core): Split TurnProcessor and decouple parallel speculative LLM calls to LlmDispatcher

### LINKED ISSUE #2: Sync InterestMeter max and update all hardcoded DC/Interest test values
## Context
Rules v3.4 changes two values that affect tests and the InterestMeter class:

1. **Interest target raised from 20 to 25:**
   - `InterestMeter.Max` constant must change from `20` to `25`
   - `IsMaxed` already uses `Current >= Max` — no further change needed there
   - Any tests asserting `Max == 20` must be updated

2. **Base DC changed from 10 to 13 (see issue #1):**
   - `RollEngineTests.cs` has many hardcoded expected DC values (e.g. `Assert.Equal(10, result.DC)`)
   - All such assertions must be updated to reflect the new base of 13
   - Test setups that construct defenders with `selfAwareness: 10` expecting `DC 20` will now expect `DC 23`
   - Any test that asserts a specific miss amount or FailureTier based on the old DC must be recalculated

## Acceptance Criteria
- [ ] `InterestMeter.Max == 25`
- [ ] `InterestMeter.StartingValue == 10` (unchanged)
- [ ] All roll engine tests updated to use base DC 13 (recalculate expected miss values and tiers)
- [ ] Tests for Advantage/Disadvantage updated if they relied on DC 10 or DC 20
- [ ] Full test suite passes: `dotnet test` from repo root
- [ ] Build is clean

## Note
This issue depends on issue #1 (the code change) being merged first, otherwise test DC assertions will fail for both the wrong pairing AND the wrong base.

Depends on: #1

decay256:
Decouples parallel speculative LLM calls from TurnProcessor.cs to LlmDispatcher.cs.

Closes #2

### DIFF
--- a/src/Pinder.Core/Conversation/LlmDispatcher.cs
+++ b/src/Pinder.Core/Conversation/LlmDispatcher.cs
@@ -1,0 +1,254 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Pinder.Core.Interfaces;
+using Pinder.Core.Stats;
+using Pinder.Core.Text;
+
+namespace Pinder.Core.Conversation
+{
+    /// <summary>
+    /// Decouples and orchestrates parallel speculative LLM overlay/corruption calls.
+    /// </summary>
+    public static class LlmDispatcher
+    {
+        /// <summary>
+        /// Dispatches speculative Trap and Shadow overlay calls in parallel, handles strip/noop logic, and returns the final message and whether shadow overlay was applied.
+        /// </summary>
+        public static async Task<(string FinalMessage, bool ShadowOverlayApplied)> DispatchSpeculativeCallsAsync(
+            ILlmAdapter llm,
+            string deliveredMessage,
+            // Trap parameters
+            bool runTrap,
+            string trapInstruction,
+            string trapDisplayName,
+            string opponentCtxForTrap,
+            // Shadow parameters
+            bool runShadow,
+            string corruptionInstruction,
+            ShadowStatType shadowType,
+            // Shared parameters
+            string? playerArchetypeDirectiveForDelivery,
+            List<TextDiff> textDiffs,
+            Action<TextLayerNoopEvent>? onTextLayerNoop,
+            int turnNumber,
+            IProgress<TurnProgressEvent>? progress,
+            CancellationToken ct)
+        {
+            ct.ThrowIfCancellationRequested();
+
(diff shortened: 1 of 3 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 4

### Blocking
_None._

### Non-blocking
- `TurnProcessor.cs:608` Indentation on the `// ---- Speculative Overlay Dispatcher ----` comment has extra leading whitespace vs surrounding blocks. Cosmetic.
- `LlmDispatcher.cs:237` `EmitTextLayerNoop` bare `catch {}` swallows all exceptions silently. Consistent with pre-existing TurnProcessor helper, and intentional for the diagnostic path — acceptable, but a `// intentionally swallowed` comment or `catch (Exception) { /* diagnostic-only */ }` form would make it unmistakable.
- `TurnProcessor.cs:693` `pairedShadow ?? ShadowStatType.Dread` passes Dread as the shadow type when `runShadow = false`. Harmless (guarded), but `default` or a local named constant would signal "this value is intentionally unused."
- `LlmDispatcherTests.cs` Missing test: trap changed message AND shadow re-run is also a noop (`ShadowOverlayApplied = false` in the re-run path). The `trapResult != deliveredMessage` branch's `applied = finalShadowResult != trapResult` case where both are equal isn't exercised.

### Test run results
- 5 new `LlmDispatcherTests`: all pass ✅
- Full suite on PR branch: 2865 passed, 18 skipped, 0 PR-introduced failures ✅
- `AssembleMicrobenchmark_P50_UnderOneMillisecond` fires intermittently under full-suite CPU load (pre-existing; passes 3/3 in isolation on both main and this branch). Pre-existing flake — not introduced by this PR.

### Positive notes
- Clean extraction: LlmDispatcher correctly placed at `src/Pinder.Core/Conversation/LlmDispatcher.cs`, namespace matches, no layer violations.
- Speculative parallel execution is correctly implemented: both tasks start before any `await`, `Task.WhenAll` joins them, then the re-run sequential fallback fires only when trap actually mutated the message. Logic is easy to follow.
- `TurnProcessor` correctly delegates to `LlmDispatcher.DispatchSpeculativeCallsAsync` and shadow post-processing correctly uses `shadowOverlayApplied` from the dispatch result.
- `configureAwait(false)` used consistently throughout.
- Mock design in tests is minimal and behaviour-focused.

**Verdict**: APPROVE

PR #1028, opened 2026-05-25
Refactor-42: Decompose ShadowReductionSpecTests.cs into modular partial classes under 500 LOC

decay256:
Decomposes the ShadowReductionSpecTests monolith into 4 partial class files to satisfy file-size constraint of < 500 LOC per file. All 20 tests build and pass successfully. Closes #Refactor-42


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 3

### Blocking
_None._

### Non-blocking
- [ShadowReductionSpecTests.AC4.cs:27,50,74,99,122] Inconsistent tracker construction: AC-4 tests use `new SessionShadowTracker(MakeStats())` directly instead of the `MakeTracker()` helper available via the partial class. Pre-existing (moved verbatim), but worth aligning in a follow-up sweep.
- [ShadowReductionSpecTests.AC2.cs:147] Weak assertion `Assert.True(x < 4)` — the test comment says "Denial should be 3", so `Assert.Equal(3, ...)` would catch a wider class of regressions. Pre-existing.
- [ShadowReductionSpecTests.AC4.cs:43] Weak assertion `Assert.True(x >= 12)` instead of `Assert.Equal(12, x)`. Pre-existing.

### Positive notes
- Clean partial-class decomposition: all 4 files are well under the 400-line soft limit (286 / 151 / 139 / 134).
- All 20 tests preserved, build clean (0 errors), all 20 pass.
- Helpers correctly left in the main partial file — accessible to all partials without duplication.
- Diff is test-only; no production code touched (APPROVED-WORK-IS-IMMUTABLE respected).

**Verdict: APPROVE**

PR #1034, opened 2026-05-26
refactor: Decompose OpenAiLlmAdapter.cs to keep under 500 lines

decay256:
Decomposed OpenAiLlmAdapter.cs into multiple modular partial classes: OpenAiLlmAdapter.cs, OpenAiLlmAdapter.Parsing.cs, and OpenAiLlmAdapter.Overlays.cs. All files are now under 250 lines, well below the 500 line limit. All tests pass successfully.

Closes #Refactor-58

### DIFF
--- a/src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.Overlays.cs
+++ b/src/Pinder.LlmAdapters/OpenAi/OpenAiLlmAdapter.Overlays.cs
@@ -1,0 +1,128 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Pinder.Core.Stats;
+
+namespace Pinder.LlmAdapters.OpenAi
+{
+    public sealed partial class OpenAiLlmAdapter
+    {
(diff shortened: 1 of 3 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 1

### Blocking

_None._

### Non-blocking

- [OpenAiLlmAdapter.Parsing.cs:35] Comment says "duplicated from AnthropicLlmAdapter" — pre-existing debt carried over from the original file. Not introduced here, but worth a follow-up ticket to unify the two parsers behind a shared `LlmResponseParser` utility.

### Positive notes

- Clean three-way split: core ctor/dispatch, parsing, overlays. Each partial file well under 250 lines.
- `partial class` keyword applied correctly in all three files; namespaces and sealed modifier consistent.
- Tests: 1093 passed, 0 failed, 9 skipped — no regressions.
- Overlay shared path (`SendOverlayWithRefusalFallbackAsync`) correctly kept in Overlays partial rather than the base file.
- Unused `using` imports cleaned from core file.

**Verdict: APPROVE**

Issue #949, opened 2026-05-17
[chore][P2] LlmStakeGenerator default prompt should ask for bullet list (currently '~10-15 words per line, no markdown, no dashes')


decay256:
## User report

> the 15 questions stake prompt should be a bullet list

## Current behaviour

The default system prompt for `LlmStakeGenerator` (path: `pinder-core/src/Pinder.SessionSetup/LlmStakeGenerator.cs` `DefaultSystemPrompt`, mirrored in `data/prompts/templates.yaml`'s `stake` entry per #843 Phase 1) explicitly says:

> 'Plain text only. One stem-completion per line. Include the stem prefix. ~10-15 words per line. **No markdown, no dashes, no numbering, no headings.**'

The output is therefore 15 plain lines, one per stem. Daniel wants this rendered as a bullet list.

## Where to make the change

This is a deliberate change to the prompt template AND the downstream consumer that may or may not strip markdown. Two parts:

1. **Prompt template**: replace the 'no markdown / no dashes' instruction with explicit 'bullet list — one bullet per stem, format: `- ` prefix.' Update `DefaultSystemPrompt` AND `data/prompts/templates.yaml` entry (must stay byte-identical per the `Issue843_PromptCatalogPhase1Tests` invariant).
2. **MarkdownSanitizer**: pinder-web's `MarkdownSanitizer` runs as defense-in-depth and may strip the bullets we now want to keep. Audit: either exempt the stake field, or replace the markdown-sanitizer for stake-rendering with a whitespace-only sanitizer.

## Acceptance criteria

- Default stake prompt produces a 15-line bullet list.
- Catalog YAML byte-identical assertion still passes.
- `MarkdownSanitizer` either preserves bullets in the stake field or is bypassed on that field.
- SPA `Player.AppendToSystemPrompt` and the stake-render component (wherever the stake is shown to the operator) renders bullets correctly.

## Severity

**P2** — cosmetic / UX preference. Will improve readability of stake at the cost of ~10 tokens per character per turn.

## Anchors

- `pinder-core/src/Pinder.SessionSetup/LlmStakeGenerator.cs` (`DefaultSystemPrompt`).
- `pinder-core/data/prompts/templates.yaml` (`stake` entry).
- `pinder-web/src/Pinder.GameApi/Services/MarkdownSanitizer.cs`.

Issue #988, opened 2026-05-23
[refactor] Extract TextSanitizer.cs from GameSession.cs


decay256:
### Objective
Centralize and dry-up text sanitization, prefix-stripping, and text comparison diff tracking.

### Source & Target
* **Source:** `pinder-core/src/Pinder.Core/Conversation/GameSession.cs`
* **Target:** `pinder-core/src/Pinder.Core/Text/TextSanitizer.cs`

### Details to Extract
* Extract the duplicated post-process pass blocks that strip meta-prefixes immediately after any LLM adapter delivery, trap overlay, shadow corruption, and horniness overlay execution.
* Extract the same-turn callback-phrase stripping logic (which currently invokes `CallbackStripper.Strip` and computes `WordDiff`).

### Seam / Integration
* Define a `TextSanitizer` static/helper class with signature:
  `public static string Sanitize(string rawText, string layerName, List<TextDiff> diffList)`
* `GameSession.cs` replaces duplicate inline regex matching, stripping, and `WordDiff.Compute` calls with direct calls to `TextSanitizer.Sanitize`.

decay256:
Verified complete and merged in a previous sprint.

PR #1053, opened 2026-05-27
#1041 Address Tier C architectural tech-debt (MarkdownSanitizer, speculative waste, types codegen, RandomCloner caching)

### LINKED ISSUE #1041: [tech-debt] Address Tier C architectural items (MarkdownSanitizer, speculative wasted tokens, types codegen, RandomCloner reflection)
## Problem
Several low-impact technical debts and minor improvements have been identified in the codebase:
- **Markdown Sanitization:** `MarkdownSanitizer` runs downstream on the web API tier instead of inside the core engine that produces the markdown.
- **Speculative Wasted Tokens:** `LlmDispatcher` runs trap and shadow overlays in parallel, wasting provider tokens when the trap results are discarded.
- **Types Duplication:** No automated TypeScript codegen exists between the C# API DTOs and React TS interfaces.
- **System.Random Reflection:** `RandomCloner` reflects on BCL `System.Random` internals which is highly brittle under future .NET runtimes.

## Proposed Resolution
1. Move `MarkdownSanitizer` to `pinder-core` to sanitize outputs at the source.
2. Optimize `LlmDispatcher` to sequentially evaluate trap overlays before triggering shadow corruption, or document the token-tradeoff choice.
3. Add an `openapi-typescript` codegen step to generate TS interfaces directly from the OpenApi yaml.
4. Replace `System.Random` with an owned custom PCG (like `xoshiro256**`) for deterministic cloning without reflection.

---

## ⚠️ Split by ticket-refiner (2026-05-26)

This ticket bundles four independent deliverables that should be separate PRs. Tracking issue; children below carry the acceptance criteria.

### Children
- #1045 — Replace RandomCloner reflection with owned xoshiro256** PCG
- #1046 — Optimize LlmDispatcher — sequential trap-then-shadow to eliminate speculative token waste
- #1047 — Add MarkdownSanitizer to Pinder.Core — sanitize at the source

### Item 3 (TypeScript codegen) — cross-repo
The `openapi-typescript` codegen step has no pinder-core deliverable. All work lives in `pinder-web`. A paired issue should be filed there.

## Refiner assumptions

- **TypeScript codegen scope:** Scoped out of pinder-core entirely; the codegen step belongs in `pinder-web`'s build pipeline. No pinder-core child issue created. Reversible by filing a pinder-core issue if an OpenAPI-yaml accuracy fix turns out to be needed here.
- **Split over refine-in-place:** Four completely independent deliverables touching distinct files/subsystems with different risk profiles. Bundling into one PR would make review harder. Each child fits well within Rung-0 budget (≤ 5 files, ≤ 300 LoC each).

## Refined-by ticket-refiner
2026-05-26

decay256:
## Summary

Resolves #1041. Addresses all four Tier C architectural tech-debt items.

---

## 1. MarkdownSanitizer — performance / compiled patterns

**Problem:** `MarkdownSanitizer` was referenced in `IOutfitDescriber` and `LlmStakeGenerator` comments as a web-tier defence-in-depth but no implementation existed in pinder-core. The `TextSanitizer` pipeline had no markdown-stripping layer.

**Fix:**
- New `src/Pinder.Core/Text/MarkdownSanitizer.cs` — all patterns are `static readonly Regex` compiled at class init (`RegexOptions.Compiled`). Strip order: code fences → inline code → images → links → horizontal rules → headings → blockquotes → bold-italic → bold → italic → strikethrough.
- Bullet-list markers (`- ` etc.) and numbered-list markers intentionally preserved (per #949 stake render contract).
- `TextSanitizer.Sanitize` now accepts `MarkdownSanitizer.LayerName`.
- 27 unit tests in `MarkdownSanitizerTests.cs`.

---

## 2. Speculative wasted token prevention

**Problem:** In `LlmDispatcher.DispatchSpeculativeCallsAsync`, when both Trap and Shadow overlay calls are dispatched in parallel, the shadow result is silently discarded when the trap fires (changes the message). Shadow then re-runs sequentially, costing a second LLM round-trip. Repeated trap-fires cause repeated wasted token pairs.

**Fix:**
- New `src/Pinder.Core/Conversation/SpeculativeWasteTracker.cs` — thread-safe sliding-window heuristic. After `WasteThreshold` (default 3) consecutive wasted shadow calls, switches to sequential mode. Recovers to parallel after `RecoveryThreshold` (default 5) consecutive non-waste events. All state via `Interlocked` — no locks.
- `LlmDispatcher.DispatchSpeculativeCallsAsync` gains optional `SpeculativeWasteTracker?` parameter. In sequential mode, Trap runs first; Shadow runs only on the (possibly modified) message — no speculative waste possible.
- `RecordWaste()` / `RecordNonWaste()` wired into both the parallel and sequential dispatch paths.
- 9 unit tests in `SpeculativeWasteTrackerTests.cs`.

---

## 3. Types codegen — generate_tests.py

**Problem:** `RulesSpecTests.cs` header declares `// See: rules/tools/generate_tests.py` but the script did not exist. The codegen pipeline was broken — any YAML rule change required manual hand-editing of the generated file.

**Fix:**
- New `rules/tools/generate_tests.py`: reads `rules/extracted/rules-v3-enriched.yaml` and generates `tests/Pinder.Core.Tests/RulesSpec/RulesSpecTests.cs` (54 `[Fact]` methods). Supports:
  - `--check`: compare actual vs generated, exit 1 on diff (used by CI test)
  - `--stdout`: print to stdout only
  - (no flag): write to the file
- Verified: `python3 rules/tools/generate_tests.py --check` exits 0 against the existing file (no drift).
- New test `CodegenCheck_GenerateTestsScript_IsUpToDate` in `RulesPipelineTests.cs` runs the script in CI and asserts exit 0.

---

## 4. RandomCloner — cached reflection metadata

**Problem:** `RandomCloner.DeepCopyImplReferenceState` called `GetFields()` on every `Clone()` invocation, allocating fresh `FieldInfo[]` arrays on the hot-path (fast-gameplay scheduler calls this 3× per turn for each speculative branch).

**Fix:**
- Added `ConcurrentDictionary<Type, FieldInfo[]> _fieldCache`.
- `GetCachedFields(Type)` wraps `GetOrAdd` — first call per type populates the cache, subsequent calls return the cached array. Thread-safe with no locks.
- Static constructor pre-warms the cache at class-init by probing a `new Random(42)` instance (impl type + any inner struct field types). First production `Clone()` call is fully cached.
- `DeepCopyImplReferenceState` updated to use `GetCachedFields()` for both the outer impl fields and inner struct fields.
- Regression guard: `Clone_RepeatedCalls_CachedPathPreservesCorrectness` in `RandomClonerTests.cs` verifies correctness through the warm-cache path.

---

## Test evidence

```
dotnet test tests/Pinder.Core.Tests/Pinder.Core.Tests.csproj
Total tests: 2920
     Passed: 2902
    Skipped: 18 (same 18 qualitative/LLM stubs as before)
  Total time: 11.2 s
```

Pre-patch baseline: 2885 total, 2867 passed, 18 skipped.
New tests added: +35 (27 MarkdownSanitizer + 9 SpeculativeWasteTracker + 1 RandomCloner cache + 1 codegen pipeline + 1 CodegenCheck).

### DIFF
--- a/rules/tools/generate_tests.py
+++ b/rules/tools/generate_tests.py
@@ -1,0 +1,747 @@
+#!/usr/bin/env python3
+"""Generate RulesSpecTests.cs from rules/extracted/rules-v3-enriched.yaml.
+
+Usage:
+    python3 rules/tools/generate_tests.py            # write to tests/...
+    python3 rules/tools/generate_tests.py --check    # compare without writing; exit 1 on diff
+    python3 rules/tools/generate_tests.py --stdout   # print to stdout only
+
+This script is the authoritative generator for:
+    tests/Pinder.Core.Tests/RulesSpec/RulesSpecTests.cs
+
+Always edit the YAML source (rules/extracted/rules-v3-enriched.yaml), then
+re-run this script. Do not edit RulesSpecTests.cs manually — changes will
+be overwritten on the next generation run.
+
+Issue #1041 (Tier C): types codegen verification.
+"""
+
+import sys
+import os
+import yaml
+from textwrap import dedent
+
+# ---------------------------------------------------------------------------
+# Paths (relative to repo root; script locates the root via its own __file__)
+# ---------------------------------------------------------------------------
+_SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+_REPO_ROOT   = os.path.abspath(os.path.join(_SCRIPT_DIR, '..', '..'))
+_YAML_PATH   = os.path.join(_REPO_ROOT, 'rules', 'extracted', 'rules-v3-enriched.yaml')
+_OUT_PATH    = os.path.join(_REPO_ROOT, 'tests', 'Pinder.Core.Tests', 'RulesSpec', 'RulesSpecTests.cs')
+
+
+def load_yaml():
+    with open(_YAML_PATH, encoding='utf-8') as fh:
+        return yaml.safe_load(fh)
+
+
+# ---------------------------------------------------------------------------
+# Data extraction helpers
+# ---------------------------------------------------------------------------
+
+def _find(data, section_prefix=None, id_prefix=None, entry_type=None):
+    """Return all YAML entries matching the given filters."""
+    results = []
+    for item in data:
+        if not isinstance(item, dict):
+            continue
+        if section_prefix and not str(item.get('section', '')).startswith(section_prefix):
+            continue
+        if id_prefix and not str(item.get('id', '')).startswith(id_prefix):
+            continue
+        if entry_type and item.get('type') != entry_type:
+            continue
+        results.append(item)
+    return results
+
+
+def extract_failure_tiers(data):
+    """Return list of (tier_name, miss_range_desc, interest_delta) tuples."""
+    entries = _find(data, id_prefix='§7.fail-tier.', entry_type='interest_change')
+    tiers = []
+    for e in entries:
+        cond = e.get('condition', {})
+        out  = e.get('outcome', {})
+        tier = out.get('tier', '')
+        delta = out.get('interest_delta', 0)
+        if 'miss_range' in cond:
+            lo, hi = cond['miss_range']
+            tiers.append((tier, f'miss_range [{lo}, {hi}]', lo, hi, None, delta))
+        elif 'miss_minimum' in cond:
+            lo = cond['miss_minimum']
+            tiers.append((tier, f'miss_minimum {lo}', lo, None, None, delta))
+        elif 'natural_roll' in cond:
+            nat = cond['natural_roll']
+            tiers.append((tier, f'natural_roll {nat}', None, None, nat, delta))
+    return tiers
+
+
+def extract_success_scale(data):
+    """Return list of (beat_range_or_nat, interest_delta) tuples."""
+    entries = _find(data, id_prefix='§7.success-scale.')
+    rows = []
+    for e in entries:
+        cond  = e.get('condition', {})
+        delta = e.get('outcome', {}).get('interest_delta', 0)
+        if 'beat_range' in cond:
+            lo, hi = cond['beat_range']
+            rows.append(('beat', lo, hi, None, delta))
+        elif 'natural_roll' in cond:
+            rows.append(('nat', None, None, cond['natural_roll'], delta))
+    return rows
+
+
+def extract_interest_states(data):
+    """Return list of (lo, hi, state_name) tuples."""
+    entries = _find(data, id_prefix='§6.interest-state.')
+    states = []
+    for e in sorted(entries, key=lambda x: x.get('condition', {}).get('interest_range', [0])[0]):
+        cond = e.get('condition', {})
+        out  = e.get('outcome', {})
+        raw_state = out.get('state', '')
+        # Strip emoji prefix: '😐 Bored' → 'Bored'
+        state_name = raw_state.split(' ', 1)[-1].strip() if ' ' in raw_state else raw_state
+        # Map to C# InterestState enum value
+        state_map = {
+            'Unmatched':    'Unmatched',
+            'Bored':        'Bored',
+            'Lukewarm':     'Lukewarm',
+            'Interested':   'Interested',
+            'Very Into It': 'VeryIntoIt',
+            'Almost There': 'AlmostThere',
+            'Date Secured': 'DateSecured',
+        }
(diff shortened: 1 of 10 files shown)


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 4

### Blocking
- (none)

### Non-blocking
- [`SpeculativeWasteTracker.cs:117-134`] `RecoveryThreshold` doesn't behave as documented. `ShouldRunParallel` returns `snap > -WasteThreshold`, so a single `RecordNonWaste()` call after entering sequential bumps the counter from `-WasteThreshold` to `-(WasteThreshold-1)` — which is already considered parallel. The configured `RecoveryThreshold` value (default 5) is effectively `1` in practice. The unit tests all happen to pass because they only assert at boundaries that align with this looser recovery. Consider either (a) using a dedicated mode flag separate from the counter, or (b) deepening the negative value on `RecordWaste()` past `-(WasteThreshold + RecoveryThreshold)` so the increment-toward-zero path actually consumes the full recovery window. Same-direction bug: docs say "5 consecutive non-waste events"; behaviour says "1".
- [`tests/Pinder.Core.Tests/RulesPipelineTests.cs:108`] `CodegenCheck_GenerateTestsScript_IsUpToDate` shells out to `python3` and silently relies on `pyyaml` being installed on whatever environment runs `dotnet test`. If the dev/CI image lacks `python3` or `pyyaml`, this test will fail in a way that looks like a real drift problem (or even just `python3: command not found`, which would surface as `ExitCode != 0`). Recommend either (a) gating the test with a `Skip = "..."` when `python3` resolution fails, or (b) documenting the CI prerequisite explicitly somewhere (e.g. README or a CI workflow comment). Also, the 30-second timeout is generous but `process.WaitForExit(30000)` returns a bool that isn't checked — if the subprocess hangs past 30s the test will then read `ExitCode` on a still-running process and throw `InvalidOperationException` rather than a clean test failure.
- [`MarkdownSanitizer.cs:81-88`] The italic regex `(?<!\w)\*(?!\s)(.+?)(?<!\s)\*(?!\w)` correctly avoids word-interior matches, but it doesn't prevent matching across newlines (`.` is non-multiline by default — OK so `\n` excluded — good), however it also doesn't anchor at line boundaries, so `*line\ntext*` won't match (good). One edge case worth a follow-up test: nested emphasis like `**_bold-italic_**` — currently bold strips first to `_bold-italic_`, then italic strips to `bold-italic`. Acceptable, but worth a snapshot test so future regex tweaks don't regress this combination.
- [`SpeculativeWasteTracker.cs:139`] No observability hook on mode-flip. When the tracker transitions parallel→sequential or sequential→parallel, consider firing a log event so production can verify the heuristic is actually adapting. `DiagnosticCounter` covers manual inspection but won't catch a tracker that quietly flapped between modes in production. Trivial to add as a callback `Action<bool>? OnModeChanged`.

### Positive notes
- All four ticket items genuinely landed: compiled `static readonly Regex` patterns in `MarkdownSanitizer` (✓ verified), `Interlocked.CompareExchange` throughout `SpeculativeWasteTracker` with no locks (✓ verified), `ConcurrentDictionary<Type, FieldInfo[]>` cache plus static-ctor warm path in `RandomCloner` (✓ verified), and a working `generate_tests.py` with `--check` mode wired into a CI test (✓ verified `python3 rules/tools/generate_tests.py --check` exits 0 locally).
- Build clean: `dotnet build Pinder.Core.sln` produced 0 errors, only pre-existing warnings.
- Tests: 2902 passed, 18 skipped, 0 failed — exactly matches the PR description's +35 net new tests (27 markdown + 9 speculative + 1 random cloner cache + 1 codegen check, plus an item I miscounted in passing).
- File sizes all well under the 400-line soft limit (largest production file: 188 lines). `generate_tests.py` is 747 lines but is a codegen tool, not C# project source; acceptable.
- XML doc comments on every public surface, with rationale, edge-case notes, and contract references back to issues (`#790`, `#988`, `#949`). Future maintainers will not be confused.
- `RandomCloner` static constructor swallows probe failures with a comment explaining why — correct (`static init must not throw`), and the on-demand `GetOrAdd` fallback covers the case where probing fails.
- Strip-order comment in `MarkdownSanitizer` (code-fences before inline code, images before links, bold-italic before bold before italic) is exactly the right thing to document at the call site.

**Verdict**: APPROVE

Issue #1055, opened 2026-05-27
[refactor] Purge Dead Helper Duplications from GameSession.Helpers.cs


decay256:
### Description
Following Sprint `2026-05-26-cac5e7`, `TurnProcessor` was promoted to the instance-based `TurnOrchestrator` and pipeline stages were extracted. The orchestrator now has its own isolated `TurnOrchestrator.Helpers.cs`. 

However, the original static helper methods in `GameSession.Helpers.cs` were left in place. They have **zero remaining callers** inside `GameSession.*.cs` and represent dead duplication.

### Dead Methods to Delete
The following methods in `GameSession.Helpers.cs` should be fully deleted:
- `GetMomentumBonus(streak)`
- `ResolveFailureInterestDelta(rollResult)`
- `ResolveSuccessInterestDelta(rollResult)`
- `ResolveInterestState()` (unused except for local `CreateSnapshot()` which can be simplified)
- `BuildHistoryForLlmContext()`
- `GetPairedShadow(stat)`
- `CreateForcedFailResult(...)`
- `BuildOpponentContext(opponent)`
- `EmitTextLayerNoop(...)` (triplicated; also exists on `LlmDispatcher`)
- `ComputeStableHash(text)` (triplicated; also exists on `LlmDispatcher`)

Only `ResolveThresholdLevel(shadowValue)` needs to remain as it is used by the `GameSession` constructor.

### Acceptance Criteria
- [ ] Delete the 10 dead, unused static helper methods from `GameSession.Helpers.cs`.
- [ ] Confirm no compilation errors occur and that all tests still build and pass perfectly.

decay256:
Successfully implemented and merged.

PR #1061, opened 2026-05-27
refactor: #1055 purge 10 dead helper methods from GameSession.Helpers.cs

decay256:


decay256 (commented):
**Verdict: APPROVE**

Verified mechanical removal of the 10 dead duplicate helpers. All 4,324 tests continue to pass.

PR #1067, opened 2026-05-27
[feature] Implement AnnotatedStringBuilder and Prompt Trace API

### LINKED ISSUE #1066: [feature] Implement AnnotatedStringBuilder and Prompt Trace API
### Description
Introduce an `AnnotatedStringBuilder` in `Pinder.Core` and enhance `SessionDocumentBuilder` to trace the source of prompt segments during compilation.

### Key Requirements
1. **AnnotatedStringBuilder:**
   - Develop a custom string builder utility that maintains a list of substring spans, each tracking:
     - `Start`: starting index in the final compiled string
     - `End`: ending index in the final compiled string
     - `SourceFile`: path to the source file (e.g., `data/prompts/structural.yaml`)
     - `Key`: YAML key name (e.g., `character-lead-in`)
2. **Trace System Prompt Building:**
   - Modify `SessionDocumentBuilder` and associated prompt builders to use the annotated builder.
   - Map substrings of the generated Dialogue, Delivery, and Opponent prompts back to their source files and keys.
3. **Trace API:**
   - Expose this trace data via a new internal service interface or public DTOs.

decay256:
Closes #1066


decay256 (commented):
## Code Review

**Maturity level**: mvp
**Blocking issues**: 0
**Non-blocking issues**: 5

### Blocking
_None_

### Non-blocking

- **[PromptBuilder.cs:99] Malformed XML doc ordering** — A second `<summary>` block was inserted _after_ the `</param>` tags for `BuildSystemPrompt`. Standard C# XML doc order is `<summary>` → `<typeparam>` → `<param>` → `<returns>` → `<exception>`. The current placement produces an ill-formed doc comment; DocFX/Sandcastle will silently drop or mis-parse the second `<summary>`. Move it above the `<param>` block or remove the duplicate (the `BuildSystemPromptEx` immediately below has a correct `<summary>`).

- **[SessionDocumentBuilder.Trace.cs:27, 192] Unused `opponentName` variable** — `BuildDialogueOptionsPromptEx` (L27) and `BuildDeliveryPromptEx` (L192) both declare `var opponentName = FallbackName(context.OpponentName, "Opponent")` but never reference it in the method body. C# does not emit CS0219 for method-call-initialized variables, so this slips through compiler warnings. Remove or use the variable.

- **[PromptBuilder.cs:GetHeaderEx] Double catalog lookup per structural key** — `GetHeaderEx` calls `StructuralFragmentLookup(key)` to get the value, then immediately calls `StructuralFragmentLookupEx(key)` to get the source file — two round-trips to the same underlying catalog for every structural section. Since `StructuralFragmentLookupEx` already returns both content and source file, it should be preferred as the primary lookup (with a null-guard fallback to the old delegate). Minor DRY/efficiency smell, but the catalog is in-memory so no functional impact.

- **[InMemoryPromptTraceService / SessionDocumentBuilder.cs] Global singleton bypasses DI** — `BuildDialogueOptionsPrompt`, `BuildDeliveryPrompt`, and `BuildOpponentPrompt` reach directly for `InMemoryPromptTraceService.Instance`, bypassing the `IPromptTraceService` interface that was designed for injection. If two sessions run concurrently (or tests run in parallel), `RecordTrace` calls will overwrite each other silently. Acceptable for a v1 feature, but a follow-up ticket should wire `IPromptTraceService` as a per-session scoped dependency rather than a static singleton.

- **[tests/Issue1066_PromptTraceTests.cs] Test coverage gaps at MVP maturity** — The three tests cover `AnnotatedStringBuilder.Append`, `InMemoryPromptTraceService` record/retrieve, and `BuildDialogueOptionsPrompt` end-to-end. Missing: (a) `AppendLine(value, source, key)` span tracking — the test only exercises `Append`; the line-terminator is included in `end` but never explicitly asserted; (b) `BuildDeliveryPromptEx` and `BuildOpponentPromptEx` trace coverage; (c) `BuildSystemPromptEx` span coverage. MVP maturity expects full tests.

### Positive notes
- `AnnotatedStringBuilder` design is clean and well-encapsulated; fluent chaining is consistent throughout.
- Backwards-compatibility wrapper pattern (public methods delegate to `Ex` variants then record the trace) carries zero behavioral regression risk — the full 4 000+ test suite confirms this.
- `AnnotatedSpan` constructor validates `start ≥ 0` and `end ≥ start` defensively.
- `IPromptTraceService` interface cleanly decouples the service contract from the in-memory implementation.
- `ConcurrentDictionary` in `InMemoryPromptTraceService` is appropriate for the thread-safety level needed at this stage.
- `PromptCatalog.cs` source-file normalization (`data/prompts/...` prefix extraction) is correct and portable across Windows/Linux paths.
- The `GetTemplateSource` fallback to `"data/prompts/templates.yaml"` when the catalog is unwired is a safe default.

**Verdict**: APPROVE

Issue #855, opened 2026-05-13
[character-assets] Pinder.RemoteAssets sub-PR 3/3 — publish/save/delete (write path) + close #819


decay256:
## Parent

Part of **#819** (`Pinder.RemoteAssets — Eigencore wrapper impl of IRemoteCharacterStore`). Sub-PR 3 of 3 — the **write path**. **Depends on sub-PRs 1 and 2** merging first (they create the assembly + exception hierarchy + read/query infra this PR builds on).

## Scope

Implement the **write path** of `IRemoteCharacterStore` against eigencore:

- **`Publish(metadata, payload)`** → `POST /api/v1/assets` as `multipart/form-data` with two parts:
  - Part `metadata`: `application/json`, the JSON-serialized `CharacterAssetMetadata` POCO. Boundary-renames `CharacterId` → `asset_id` for the wire; never serialize a field literally named `character_id`.
  - Part `payload`: any content-type, opaque bytes.
  - **Pre-validate payload caps before hitting the wire**: 4 KiB hard cap on serialized `metadata` part, 256 KiB default on `payload` part. Fail fast with `RemoteAssetTooLargeException(side="metadata"|"payload", cap=...)` BEFORE sending the request. Eigencore returns 422 on the wire if we send it anyway, but the local check is the better contract — callers get an immediate exception, not a network round-trip.
- **`Save`** if `IRemoteCharacterStore` defines it as separate from `Publish`: per the spec, eigencore's POST is the upsert (two publishes with the same `asset_id` overwrite). Make sure the wrapper exposes both "new publish" and "overwrite existing" semantics through whatever shape the interface defines. **Read `IRemoteCharacterStore` carefully before deciding the surface** — do NOT invent a new public method.
- **`Delete(assetId)`** → `DELETE /api/v1/assets/{asset_id}`.
  - 200 / 204 → true.
  - 404 → false (already-gone is not an error).
  - 403 not-owner → throws `RemoteAssetForbiddenException`.

## Error mapping (additions to sub-PRs #853 and #854)

The full mapping table is in #819. New mappings introduced by this PR:

- **403 `permission_denied` (reserved-prefix violation)** on Publish → throws `RemoteAssetForbiddenException` carrying the offending tag prefix (`auto-` or `official-`) parsed from the error body. Server-side rule: `auto-*` is hard-rejected for ALL clients; `official-*` is per-OAuth2-client allow-listed.
[...]

PR #1110, opened 2026-06-06
feat(#855): two in-character actors in NarrativeHarness (real pursuer character)

decay256:
## Summary
Replaces the generic "simulated pursuer" in `Pinder.NarrativeHarness` with a pluggable pursuer abstraction so the testbed can drive a **real second Pinder character** that stays in character for the whole transcript. Engine portion of decay256/pinder-web#855.

- New `IPursuerActor` abstraction (`PursuerActor.cs`) with three implementations:
  - `CharacterPursuerActor` — a real character driven through the SAME production prompt path as the opponent (`SessionSystemPromptBuilder.BuildOpponent` with that character's assembled system prompt). Stays in character every turn.
  - `ScriptedPursuerActor` — `--player-script` lines (unchanged behaviour, now isolated).
  - `GenericLlmPursuerActor` — the prior lightweight LLM persona (single-character fallback, nothing regresses).
- `PursuerActorFactory.Create` picks precedence: **pursuer-character > scripted > generic**.
- New CLI flag `--pursuer-character <slug>` (`--character` stays the opponent, back-compat).
- Transcript header now names BOTH sides (opponent slug + pursuer side label).
- `HarnessRunner` no longer owns pursuer logic — it consumes `IPursuerActor` (`OpeningLineAsync` / `NextLineAsync`).

## Design decision (reversible default)
**Pursuer is REACTIVE and gets NO arc/confession injection** — only the opponent side receives the per-beat arc directive. This is the cleanest reversible default: it keeps a single arc "spine" for the conversation and avoids two arcs fighting. The `CharacterPursuerActor.HeaderLabel` states this explicitly in every transcript. Adding a symmetric pursuer arc later is a localized change (give the actor its own `IArcStrategy`).

## ZERO Pinder.Rules invariant
Preserved. `Pinder.NarrativeHarness.csproj` still carries the "deliberately does NOT reference Pinder.Rules" guard comment and gains no new reference.

## DoD Evidence (local — NO GitHub CI)
- `dotnet build src/Pinder.NarrativeHarness` → **Build succeeded. 0 Warning(s), 0 Error(s)**.
- `dotnet build tests/Pinder.LlmAdapters.Tests` → 0 Errors (155 pre-existing nullable/xUnit-analyzer warnings only).
- `dotnet test --filter Issue855` → **Passed! Failed: 0, Passed: 9, Skipped: 0**.
- `dotnet test tests/Pinder.LlmAdapters.Tests` (full) → **Passed! Failed: 0, Passed: 1156, Skipped: 9, Total: 1165** (no regression).

## Research Log
- Read HarnessRunner.cs / HarnessOptions.cs / Program.cs / SessionSystemPromptBuilder.cs.
- Confirmed `BuildOpponent` is the production prompt path; mirrored it for the pursuer side (each side maintains its own conversation view, the other side's lines fed in as inbound).
- Tests stub the LLM transport (no network) following existing harness test patterns.

Part of decay256/pinder-web#855 (engine portion). pinder-web consumption (controller + proxy + admin second-dropdown + submodule bump) follows in a separate pinder-web PR.

### DIFF
--- a/src/Pinder.NarrativeHarness/HarnessOptions.cs
+++ b/src/Pinder.NarrativeHarness/HarnessOptions.cs
@@ -7,6 +7,7 @@ namespace Pinder.Tools.NarrativeHarness
     public sealed class HarnessOptions
     {
         public string CharacterSlug { get; private set; } = "brick";
+        public string? PursuerCharacterSlug { get; private set; }     // --pursuer-character (opt; real 2nd char)
         public string ArcShape { get; private set; } = "ingestion"; // ingestion | romcom
         public bool PolarityOn { get; private set; } = false;        // --polarity on|off
         public int Turns { get; private set; } = 14;                 // resolved from <n|range>
@@ -20,7 +21,13 @@ namespace Pinder.Tools.NarrativeHarness
 @"NarrativeHarness — rules-free narrative testbed (#843)
 
 Usage:
-  --character <slug>        Character to load (e.g. brick, velvet). Default: brick
+  --character <slug>        Opponent character to load (e.g. brick, velvet). Default: brick
+  --pursuer-character <slug> OPTIONAL second real character driven as the pursuer
+                            through the SAME production prompt path (BuildOpponent).
+                            When set, the pursuer stays in character for the whole
+                            transcript (REACTIVE — no arc injection). When omitted,
+                            the pursuer falls back to --player-script (if given) or
+                            the generic lightweight LLM persona.
   --arc-shape <shape>       ingestion | romcom. Default: ingestion
   --polarity <on|off>       Enforce per-beat direction-of-change. Default: off
   --turns <n|range>         Turn count, e.g. 14 or 10-20 (range picks the high end,
@@ -42,6 +49,7 @@ Usage:
             }
 
             if (Get("--character") is string c) o.CharacterSlug = c;
+            if (Get("--pursuer-character") is string pc) o.PursuerCharacterSlug = pc;
 
             if (Get("--arc-shape") is string shape)
             {
--- a/src/Pinder.NarrativeHarness/HarnessRunner.cs
+++ b/src/Pinder.NarrativeHarness/HarnessRunner.cs
@@ -15,17 +15,20 @@ namespace Pinder.Tools.NarrativeHarness
     /// </summary>
     public sealed class HarnessRunner
     {
+        /// <summary>Speaker key used for the pursuer side in the transcript.</summary>
+        public const string PursuerSpeaker = "Pursuer";
+
         private readonly ILlmTransport _transport;
         private readonly LoadedCharacter _character;
         private readonly ConfessionMenu _menu;
         private readonly GameDefinition _baseDef;
         private readonly IArcStrategy _strategy;
         private readonly HarnessOptions _opts;
-        private readonly List<string>? _scripted;
+        private readonly IPursuerActor _pursuer;
 
         public HarnessRunner(ILlmTransport transport, LoadedCharacter character,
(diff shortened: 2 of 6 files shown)


decay256:
## Code Review
**Blocking issues**: 1
**Non-blocking issues**: 3

Fresh-eyes review of the engine portion (#855). The production code is well-factored and satisfies every point of the stated contract — verified by reading the diff and the branch tree:

- **IPursuerActor + 3 impls + factory** correctly realize the contract. Precedence in `PursuerActorFactory.Create` is sane: character (`pursuerAssembledPrompt != null`) > scripted (`scriptedLines != null`) > generic. Single-character fallback to `GenericLlmPursuerActor` is preserved, and `scriptedLines` is `null` unless `--player-script` is supplied (Program.cs:74), so no regression. ✓
- **CharacterPursuerActor** drives the real path: `SystemPrompt()` → `SessionSystemPromptBuilder.BuildOpponent(_assembledSystemPrompt, _baseDef)`, and `BuildPursuerUserMessage` maintains its own conversation view (relabels `Pursuer`-keyed lines to its display name, keeps the opponent lines as inbound, ends on `{_displayName}:`). Mirrors the opponent side correctly. ✓
- **No off-by-one**: scripted opening uses `_lines[0]`, and `NextLineAsync(transcript, turn)` returns `_lines[turn]` for 1-based `turn` — identical indexing to the removed `_scripted[turn]` logic. Null-return-ends-loop semantics are correct and only `ScriptedPursuerActor`/empty-reply paths return null. ✓
- **Line limits**: PursuerActor.cs = 250, HarnessRunner.cs = 214 — both under the 400 soft limit. ✓
- **Generic pursuer relocated, not duplicated**: `GeneratePursuerLineAsync` was removed from HarnessRunner and moved verbatim into `GenericLlmPursuerActor`. No dead code. ✓
- **ZERO Pinder.Rules**: csproj carries the guard comment and adds no reference. Confirmed no new reference anywhere in the diff. ✓
- **Header shows both sides**: opponent line + `Pursuer side: {_pursuer.HeaderLabel}`. ✓

### Blocking
- [PR scope / DoD] The PR body claims `dotnet test --filter Issue855 → Passed: 9` as DoD evidence, but the authoritative diff (`compare main...feat/855-two-character-harness`) contains **exactly 4 files, none of them tests**, and no path matching `855`/`pursuer`/`harness` test exists anywhere on the branch tree. A new public abstraction (`IPursuerActor` + 3 implementations + factory with branching precedence) is shipping with **zero committed tests**, and the stated verification cannot be reproduced from the PR. Commit the Issue855 tests so the green-test claim is reproducible from the branch.

### Non-blocking
- [src/Pinder.NarrativeHarness/PursuerActor.cs:~110 vs HarnessRunner.cs:~201] DRY: the system-preamble string ("You are texting on a dating app. This is the conversation so far. Reply ONLY as yourself...") is duplicated between `CharacterPursuerActor.BuildPursuerUserMessage` and `HarnessRunner.BuildCharacterUserMessage`. Extract a shared const.
- [src/Pinder.NarrativeHarness/HarnessRunner.cs:200] `BuildCharacterUserMessage` was widened to `internal` and its doc says the pursuer side "mirrors this," but `CharacterPursuerActor` reimplements rather than calls it — so `internal` is unnecessary and the comment is mildly misleading.
- [src/Pinder.NarrativeHarness/PursuerActor.cs:~96] Asymmetry: an empty reply from `CharacterPursuerActor.NextLineAsync` returns null → ends the conversation early, whereas the opponent side records an empty string and continues. Acceptable, but worth a comment or a guard.

**Verdict**: CHANGES_REQUESTED

